PR #3, opened 2023-12-04
feat:  ✨ Pull in base OTel WebSDK

pkanal:
This PR pulls in the code that will eventually become the OpenTelemetry WebSDK. For now we're using all the code from the [draft PR](https://github.com/open-telemetry/opentelemetry-js/pull/4325) that is up so it is subject to change, we're not worried about it changing too much because it is very similar to the NodeSDK which is a released package. 

The main new piece of this SDK is the `SpanIdSpanProcessor`, ~I have chosen to leave this component out for now because it is pretty unstable and undefined and [still depends on `Buffer` which is a Node-only API](https://github.com/open-telemetry/opentelemetry-js/pull/4325/files#diff-27a5bf158857e4404f443ccb5b11e07c4486203a76f9953e1316167cf5723f01R22). We can pull this in as it becomes more stable as it will be an additive change.~ Nevermind, I figured out how to make it browser friendly because there was prior art in upstream OTel for this with trace and span IDs, this is likely what it will look like so I feel comfortable pulling it in for now.

- Pull in WebSDK code and types
- The distro will emit `session.id` 🎉 
- Update example to use the WebSDK
- Update example to emit traces

1. Build the project, in the root directory run: `npm run build`
2. Change directories to the example app: `cd examples/hello-world-web`
3. Build the example app: `npm run build`
4. Run the example app: `npm start`
5. Go to http://localhost:3000 in the browser
6. Open the console tab, you should see OTel debug messages as well as document load traces logged to the console.

### DIFF
--- a/examples/hello-world-web/index.js
+++ b/examples/hello-world-web/index.js
@@ -1,9 +1,17 @@
 import { WebSDK } from '@honeycombio/opentelemetry-web';
 import { DiagConsoleLogger, DiagLogLevel, diag } from '@opentelemetry/api';
+import { getWebAutoInstrumentations } from '@opentelemetry/auto-instrumentations-web';
+import { ConsoleSpanExporter } from '@opentelemetry/sdk-trace-base';
+
 const main = () => {
-  console.log('Hello from the example app');
+  // Set OTel to log in Debug mode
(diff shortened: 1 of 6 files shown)


MikeGoldsmith (approved):
🎉

MikeGoldsmith (approved):

PR #8, opened 2023-12-07
feat: ✨ Add browser smoke tests

pkanal:
Adds basic browser smoke tests that test whether the OTel API is initialized and that traces are produced. This is not running in CI yet.

- Add Cypress as a dependency
- Configure Cypress tests to check console output for whether API is initialized and traces are produced
- Added Docker config to run smoke tests with one command

- Check out this branch
- Make sure you have docker installed
- Run `npm run test:smoke`


mjayaram (approved):
thank you!!

PR #22, opened 2024-01-05
feat: ✨ Endpoint & API Key config

### LINKED ISSUE #9: Add Honeycomb config options
- Add config options according to the [Honeycomb distro spec](https://github.com/honeycombio/specs/blob/main/specs/otel-sdk-distro.md).

**Open Questions/Considerations**
- Does OTel web support metrics? Do we need to support metrics here?
- No need for environment variable support for now (OTel web does not support env vars)
- This ticket is to extend the `WebSDKConfiguration`, if there are certain attributes that are not hooked up yet (e.g. debug, sample rate etc.) that's ok

pkanal:
- Closes #9 

- Add Honeycomb specific config option types
- Wire up endpoint and api key config 

Out of scope(for this PR):
- Baggage Span Processor
- Resource attributes
- Sample rate
- Debug mode
- Local visualizations
- Environment variable support

- Run the example app with an API key and see traces in Honeycomb


kentquirk (commented):
I added some comments, but feel free to ignore them if I'm missing the point.

kentquirk (approved):
LGTM!

PR #30, opened 2024-01-23
feat: ✨ add entry page attributes to resource

### LINKED ISSUE #25: Add landing page attributes
As a follow-up to #10, create a new resource with information about the page's initial url: 
- full url

as well the individual bits of the url broken out into individual fields:
- search params
- hash
- page
- hostname

ahrbnsn:
- Closes #25 

Adds  the following fields to the resource:
`entry_page.url` 
`entry_page.path` 
`entry_page.search` 
`entry_page.hash` 
`entry_page.hostname`

Confirm the new fields are sent to honeycomb, and accurately reflect the initial page location

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -1,0 +1,69 @@
+# Local Development
+
+## Prerequisites
+
+**Required:**
+
+- Node.js (minimum version declared in package.json)
+
+**Recommended:**
+
+- VSCode Plugins:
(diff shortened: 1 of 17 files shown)


JamieDanielson (approved):
Left a note about adding url sanitization but that can be a follow-up, this works

pkanal (approved):

PR #31, opened 2024-01-23
maint: 🔧  add `mergeResources` util

ahrbnsn:
Adds utility function to support merging an array of resources into a singular resource

## Why?
For ergonomics & to support creating small, single-purpose resource-builders: see #30

Moves the overhead of merging resources out of the individual configuration builders & into one centralized location: pass in an array of resources, get a singular resource w/ all the available attributes back


JamieDanielson (commented):
I think this looks good, left 2 questions in the tests. The one is a nit for clarifying what we're testing... but the other I'm not sure. We may need some deduplication logic for Resources with identical keys

JamieDanielson (approved):

PR #33, opened 2024-01-23
feat: add debug mode

### LINKED ISSUE #14: Add debug options
When the `debug` config option is set to true:
- Set the standard OTel logging level to `debug`
- Log config options

JamieDanielson:

- Closes #14 

- add configureDebug function. When debug: true is set, configureDebug creates a `DiagConsoleLogger` with Debug level and also prints HoneycombOptions to console.

Set `debug: true` in SDK config and see both the output expected from the previously enabled DiagConsoleLogger, as well as configured options for the Honeycomb SDK.

for configured options of key, serviceName, endpoint (without path):
debug-for-configured-options

for missing options (is this excessive?):
debug-for-missing-options

~Note I'm not sure why the smoke tests are failing (local) now 🤔  I tried updating but may need an extra set of eyes on that, maybe as a follow-up? Confirmed the instrumentation is showing in the console from manually running the app.~
**Update**: It looks like it doesn't work because the DiagConsoleLogger only logs output if it can successfully export - the app defaults to sending to Honeycomb without a valid key, so nothing gets logged to the console. By adding a ConsoleSpanExporter, I can confirm the smoke tests pass. I'll consider that for updating the smoke tests further.

### DIFF
--- a/cypress/e2e/spec.cy.ts
+++ b/cypress/e2e/spec.cy.ts
@@ -20,6 +20,19 @@ describe('Smoke Tests', () => {
     );
   });
 
+  it('logs honeycomb config with debug enabled', () => {
+    cy.visit('http://localhost:3000', {
+      onBeforeLoad(win) {
+        cy.stub(win.console, 'debug').as('consoleDebug');
+      },
+    });
+
+    cy.get('@consoleDebug').should(
+      'be.calledWithMatch',
+      'Honeycomb Web SDK Debug Mode Enabled',
+    );
+  });
+
   it('logs document load traces', () => {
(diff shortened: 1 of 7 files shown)


pkanal (commented):
Talked this through synchronously and we decided to switch to only logging out final computed values that are the most relevant to users. E.g. instead of logging out `endpoint` and `tracesEndpoint`, we'll log out something more user friendly like "Your traces are being sent to <x>".

ahrbnsn (approved):
🎉

pkanal (approved):

Issue #24, opened 2024-01-19
Public API method for adding resource attributes


wnstn:
**Is your feature request related to a problem? Please describe.**
One of the big differentiators for Honeycomb RUM vs other RUM products is our support for wide-data. This will allow users to filter down metrics to the constraints they care about (customers, users, page route owners, etc). Since we will be selling this as a key benefit, we should make it really easy to add attributes that help with that via resource attributes.

**Describe the solution you'd like**
Ideally a customer can import a method into whatever file has the info they want to attach to all spans and pass in a JS object to that method, and our SDK takes care of attaching all K/Vs in the object as resource attributes.

In an even more ideal world, there's a way to delay span processing until all resource attributes are present, so that devs can attach the attributes they need even if they are not available at the point of SDK instantiation.

**Describe alternatives you've considered**
N/A

**Additional context**
I think I've covered everything here, but this request is mostly making sure we think about the DevEx of this common usage pattern.

PR #34, opened 2024-01-24
feat: ✨ add support for plain objects when initializing resource

ahrbnsn:
Addresses a portion of  #24 — does not add support for async adding of fields, but adds a new field to the honeycomb sdk configuration which'll let folks pass in an object w/ static attributes & get those merged into the resource attributes

Add a custom resource to the example app's config:
```js
# examples/hello-world-web/index.js

const main = () => {
  // Set OTel to log in Debug mode
  diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.DEBUG);

  // Initialize base OTel WebSDK
    apiKey: 'api-key-goes-here',
    serviceName: 'web-distro',
    instrumentations: [getWebAutoInstrumentations()], // add auto-instrumentation
    resourceAttributes: { custom_field: 'really-great' },
  });
  sdk.start();
};

main();
```

Expected: the resource includes both your custom field(s) as well as `honeycomb.distro.version` & `honeycomb.distro.runtime_version` fields


ahrbnsn:
> I'm generally in favour of this! I think the main note I have is to change the name of this to `resourceAttributes`. 
🎉 

> Also if they user does provide an actual `Resource` instance, we should account for that being merged in as well. I think this might already happen but I just want to confirm!

Right now, if folks have a bonafide otel resource they want, they'll keep passing it in under the `resource` key. This adds a new config option to supplement the base `WebSDKConfiguration` - got a resource? otel  standard-issue config handles that. have a bunch of key/values? add em to the new `resourceAttributes` option & we'll make sure they're included

This config would result in an end resource w/ both `'another_field: 'heck yes'` & `custom_field: 'really-great'`:
```
    apiKey: 'api-key-goes-here',
    serviceName: 'web-distro',
    instrumentations: [getWebAutoInstrumentations()],
   // two ways to add resource fields; instantiating a new Resource or passing in an option
    resource: new Resource({ another_field: 'heck yes' }),
    resourceAttributes: { custom_field: 'really-great' },
  });

```

pkanal (approved):
I'm excited about this API, this means that users don't have to install the resources package on top of our web distro to add some basic resource fields! 🎉 @jessitron would love to get your thoughts on adding this and whether we're overlooking any footguns by adding this in.

src/types.ts
```diff
@@ -83,6 +88,9 @@ export interface HoneycombOptions extends Partial<WebSDKConfiguration> {
    */
   debug?: boolean;
 
+  /** Any additional attributes to add to the resource */
+  resourceAttributes?: ResourceAttributes;
```
JamieDanielson, line 92: I like that we're naming it `resourceAttributes` because that's what it is, but I also acknowledge your earlier point about people maybe not understanding what it means. Maybe we can spiff up the comment a bit to say "`Any additional attributes to be included on all spans in a trace`" or something to that effect?

test/honeycomb-otel-sdk.test.ts
```diff
@@ -8,20 +8,40 @@ test('it should extend the OTel WebSDK', () => {
   expect(honeycomb).toBeInstanceOf(WebSDK);
 });
 
-/* This test relies on `getResourceAttributes`, a method not
+/* These test relies on `getResourceAttributes`, a method not
```
JamieDanielson, line 11: if we're adjusting anyway...
```suggestion
/* These tests rely on `getResourceAttributes`, a method not
```

JamieDanielson (approved):
This looks great and nice to interact with. Added a couple small suggestions around wordsmithing but generally no blocker from me!

JamieDanielson:
Oh! Actually it would be great to add this usage to the example app as you've proposed in the PR description, though that can be a follow-up PR.

src/types.ts
cartermp: Maybe "included on all data"? Since if this SDK ever starts emitting logs or metrics that'll still be true.

PR #45, opened 2024-01-26
feat: ✨ Browser attributes span processor (but also so much more)

### LINKED ISSUE #11: Add Browser attributes span processor
Add a span processor for browser attributes that might change for each span (e.g. window.location). See examples from [semantic conventions](https://github.com/scheler/opentelemetry-specification/blob/browser-events/specification/logs/events/semantic_conventions/browser.md#webvital).

pkanal:
This PR adds the `BrowserAttributesSpanProcessor` which adds rich browser attributes to every span. Attributes like the page location and screen height and width might change during the lifecycle of a session so they are better captured through a Span Processor rather than on the Resource.

- Closes #11 

- `BrowserAttributesSpanProcessor` class that adds attributes to each span
- `CompositeSpanProcessor` class that combines multiple span processors together since the base SDK does not support multiple span processors
- Span processor builder that configures the `BatchSpanProcessor` with the exporter as well as the `BrowserAttributesSpanProcessor` and an optional user provided span processor
- Had to switch the exporter being configured through this span processor builder since the base SDK does not support configuring an exporter through the `traceExporter` option if a `spanProcessor` is provided.

- Run the test app and check for the additional browser attributes
- The data should still be exported to Honeycomb

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ The SDK adds these fields to all telemetry:
 | browser.platform | stable | static | [NavigatorUAData: platform](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/platform) | "Windows" |
 | browser.mobile | stable | static | [NavigatorUAData: mobile](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/mobile) | true |
 | browser.language | stable | static | [Navigator: language](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language) | "fr-FR" |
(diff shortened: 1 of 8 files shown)


JamieDanielson (commented):
Main suggestion is regarding properties being used for width and height... otherwise on the whole this looks great!

JamieDanielson (changes requested):
Setting as Request Changes for the browser width/height suggestion

ahrbnsn (approved):

JamieDanielson (approved):
This is so great! We may have some more readme updates but I'm fine doing that as a follow-up too

PR #49, opened 2024-01-26
fix: downgrade otel packages while awaiting fix

JamieDanielson:

- error when installing latest with auto-instrumentations-web

- downgrade otel packages to known-good versions, keeping them in sync as 1.18.1 with 0.45.1 
- update readme with current versions

installing and building example app doesn't cause errors


pkanal (approved):
🎉

Issue #17, opened 2024-01-02
[ci] Connect smoke tests to CI


pkanal:
Connect cypress tests to CI

PR #50, opened 2024-01-26
maint: add mini smoke test to CI

JamieDanielson:

- ensure apps build properly on every commit

- add mini smoke test to CI that just ensures the app can build and run without error

run these locally and it should build fine, change the auto-instrumentations-web version to 0.35 and see it error. Actually we should see this fail in dependabots too until the fix comes out for auto-instrumentations.

Note this is a temporary stopgap to catch errors before we finish implementing #17

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,6 +64,21 @@ jobs:
       - store_artifacts:
           path: ./reports/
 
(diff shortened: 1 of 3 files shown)


ahrbnsn (approved):
🥇

PR #55, opened 2024-01-29
maint: build out smoke tests and connect to CI

### LINKED ISSUE #17: [ci] Connect smoke tests to CI

JamieDanielson:

- Closes #17 

- Add smoke tests using BATS, similar to our other distros. This primarily allows for re-using smoke test logic independent of language, as it inspects and asserts against telemetry collected by the Collector and exported to a file.
- Update CircleCI config to run smoke tests in every push, similar to lint/build/test
- Add Makefile for easier smoke testing. This doesn't appear to be necessary for local smoke testing, as similar scripts in the `package.json` also work locally. CI didn't seem to work as well, hence the decision to keep the Makefile.
- Update `DEVELOPING.md` with some details around smoke tests
- Update example to send to collector at localhost, without API key. This may be the most "controversial" and may warrant having a different example app - open to ideas. If we keep an API key we have end up with CORS errors for a "With-Credentials" header. 

See the [results in Circle](https://app.circleci.com/pipelines/github/honeycombio/honeycomb-opentelemetry-web/113/workflows/546edcba-4d22-4dd9-8fc3-114170373f0f/jobs/435/artifacts), including actual assertions like `Agent includes browser attributes` as well as telemetry output stored from the collector file export.


pkanal (approved):
This is so great! Ship once cleanup is done :D

PR #58, opened 2024-01-31
rel: prep release 0.0.1 alpha

JamieDanielson:

- lack of a published release!

- update changelog for first alpha release
- update readme with current versions of otel
- note: version was already set to 0.0.1, so no changes needed there

- after merging we should be able to tag and release the first alpha version


pkanal (approved):
🎉

PR #60, opened 2024-01-31
fix(ci): update publish steps to depend on smoke tests

### LINKED ISSUE #59: release should be blocked on smoke test, not parallel
**Versions**

0.0.1

**Steps to reproduce**

1. follow steps to publish/release, see smoke test in parallel to release instead of blocking

**Additional context**

we should run smoke tests and ensure they pass before publishing

pkanal:
- Closes #59 

Currently, [smoke tests run in parallel](https://app.circleci.com/pipelines/github/honeycombio/honeycomb-opentelemetry-web/136/workflows/492c3294-d552-4f4f-9cbc-ee6685ace476) to our publish steps. We want publish steps to depend on smoke tests so that if they fail, a package does not get published.

- Change publish steps to depend on smoke tests in CI config.

- This will only be verifiable on our next release. This is really similar to our setup in the [Node distro](https://github.com/honeycombio/honeycomb-opentelemetry-node/blob/main/.circleci/config.yml#L145).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,8 +138,10 @@ workflows:
           context: Honeycomb Secrets for Public Repos
           requires:
             - build
+            - smoke_test
       - publish_npm:
           <<: *filters_publish
           context: Honeycomb Secrets for Public Repos
           requires:
             - build
+            - smoke_test


ahrbnsn (approved):
🚀

PR #61, opened 2024-02-01
docs: web vitals config & event attributes

ahrbnsn:
Adds documentation for proposed web vitals config & span attributes for each vital: to kickstart discussion & align prior to writing the actual implementation code. living that [readme-driven-development](https://tom.preston-werner.com/2010/08/23/readme-driven-development.html) dream

[Formatted Web Vitals readme](https://github.com/honeycombio/honeycomb-opentelemetry-web/blob/ahr/vitals-readme/docs/web-vitals.md)

### Event Fields
**Namespacing**
Moves most of the individual attributes out of the `web_vital.` namespace & into the respective metric's acronym, `cls.value | lcp.value | inp.value | fcp.value | ttfb.value`.

Benefits to namespacing under the individual acronoym instead of a shared `web_vital`: when adding filters or group-bys, I've found it nice to have confidence I'm using a field that's actually relevant for the specific metric I'm digging into 

**Why not keep the web_vital namespace & have nested keys, a la `web_vital.cls.largest_shift_entry`**? We could do that! But for brevity's sake, both in UI display (long field names are a bear) as well as saving folks those extra key strokes. I reckon each vital has enough custom fields to warrant top-level namespace

Universal fields on all web vitals: `web_vital.name`, `web_vital.rating`, `web_vital.id`

For metrics which include reference to a specific element — largest contentful paint, cumulative layout score, interaction to next paint — that element is always included in `[metric].element`, instead of a metric-specific key (`inp.element` instead of `inp.event_target`, `cls.element` instead of `cls.largest_shift_target`. `lcp.element` remains `lcp.element`). Although we do lose specific signal around _why_ the element is being referenced, I think there's benefit to folks only needing to remember one field name. Each metric has only one specific, metric-related reason for referencing an element; I don't think we're losing too much signal by shedding the specificity in the key.

**Element attribution**
Adds a config field, `elementDataAttribute`, to let folks set a [custom data attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*) we'll look for when determining the element id. The web-vitals library uses the css selector, which is not always legible or consistent between builds (our telemetry has a bunch of `inp.event_target` that look like `div>div.css-bncyh0>div.css-1diydk0>div.fs-unmask.css-tcu5wm>a.css-1otlhad>h4.css-q2y3yl.css-bgrgbw`—good luck figuring out what _that_ corresponds to).

The readme right now has the same field, `[metric].element`, reference either the custom data attribute, if available, _or_ the web-vitals provided selector. I think there's arguably value in including both the data attribute (when available) _and_ the selector in individual fields.
Some options:

1. one field, `[metric].element`: data-attribute || selector
2. two distinct fields: `[metric].element.attribute`: data attribute `[metric].element.selecor`: the selector
   3). all the fields: `[metric].element`: data-attribute || selector, `[metric].element.attribute`: data attribute `[metric].element.selecor`: the selector

Option 3 could be handled by derived columns, but I like the idea of providing easily-queryable fields from jump.

**Event Timestamp & Duration**
When the metric has fields which correspond to an event timestamp & duration, use those values

CLS: Event time: `performance.timeOrigin + attribution.largestShiftTime`. duration n/a
LCP: Event time: `performance.timeOrigin`. Duration: `metric.value`
INP: Event time: `attribution.eventTime`. Duration: `metric.value`
FCP: Event time: `performance.timeOrigin`. Duration: `metric.value`
TTFB: Event time: `performance.timeOrigin`. Duration: `metric.value`

### Config Options
Allow disabling web vitals collection; accept custom event handlers for each of the web vitals as an escape hatch when folks need more control; allow specification of a custom data attribute for element attribution; option to keep the `[metric].value` in the `web_vital` namespace (ie `inp.value` => `web_vital.value`)

### DIFF
--- a/docs/web-vitals.md
+++ b/docs/web-vitals.md
@@ -2,13 +2,13 @@
 
 By default, Honeycomb includes events for all web vitals except first-input-delay. First-input-delay is set to be replaced by interaction-to-next-paint as a core web vital in March 2024.
 
-## Metric Attributes
+## Vitals Attributes
 
-All metrics have the following attributes: `web_vital.name`, `web_vital.rating`, `web_vital.id`
+All vitals have the following attributes: `web_vital.name`, `web_vital.rating`, `web_vital.id`
 
 ```
 /**
- * The name of the metric (in acronym form).
+ * The name of the vital (in acronym form).
  */
 "web_vital.name": 'CLS' | 'FCP' | 'INP' | 'LCP' | 'TTFB';
 
@@ -24,7 +24,7 @@ All metrics have the following attributes: `web_vital.name`, `web_vital.rating`,
  * metric instance, or to group multiple deltas together and calculate a
  * total. It can also be used to differentiate multiple different metric
  * instances sent from the same page, which can happen if the page is
- * restored from the back/forward cache (in that case new metrics object
+ * restored from the back/forward cache (in that case new vital object
  * get created).
  */
 
--- a/docs/web-vitals.md
+++ b/docs/web-vitals.md
@@ -1,6 +1,6 @@
 # Web Vitals
 
-By default, Honeycomb includes events for all web vitals except first-input-delay. First-input-delay is set to be replaced by interaction-to-next-paint as a core web vital in March 2024.
+By default, Honeycomb includes events for all web vitals except [first-input-delay](https://web.dev/articles/fid). First-input-delay is set to be replaced by interaction-to-next-paint as a core web vital in March 2024.
 
 ## Vitals Attributes
 
--- a/docs/web-vitals.md
+++ b/docs/web-vitals.md
@@ -32,7 +32,7 @@ All metrics have the following attributes: `web_vital.name`, `web_vital.rating`,
 
 ```
 
-### Cumulative Layout Score attributes
+### [Cumulative Layout Score](https://web.dev/articles/cls) attributes
 
 ```
 /* Shared fields */
@@ -84,7 +84,7 @@ All metrics have the following attributes: `web_vital.name`, `web_vital.rating`,
 
 The event time is when the single largest layout shift contributing to the page's CLS score occurred. (performance.timeOrigin + metric.largestShiftTime)
 
-### Largest Contentful Paint attributes
+### [Largest Contentful Paint](https://web.dev/articles/lcp) attributes
 
 ```
 /** Shared fields */
@@ -141,7 +141,7 @@ The event time is when the single largest layout shift contributing to the page'
 
 The event time is the start of the page load (performance.timeOrigin), the duration is equivalent to "lcp.value"
 
-### Interaction to Next Paint attributes
+### [Interaction to Next Paint](https://web.dev/articles/optimize-inp) attributes
 
 ```
 /** Shared fields */
@@ -182,7 +182,7 @@ The event time is the start of the page load (performance.timeOrigin), the durat
 
 The event time is equal to the time the interaction began {[TODO] `attribution.eventTime + performance.timeOrigin`? or just `attribution.eventTime`}, the duration is equal to `inp.value`
 
-### First Contentful Paint attributes
+### [First Contentful Paint](https://web.dev/articles/fcp) attributes
 
 ```
 /** Shared fields */
@@ -216,10 +216,9 @@ The event time is equal to the time the interaction began {[TODO] `attribution.e
 
 The event time is equal to the start of the page load, the duration is equal to `fcp.value`
 
-### Time to First Byte attributes
+### [Time to First Byte](https://web.dev/articles/ttfb) attributes
 
 ```
-
 /** Shared fields */
 "web_vital.name": "TTFB",
 "web_vital.rating": "good" | "needs-improvement" | "poor",
--- a/docs/web-vitals.md
+++ b/docs/web-vitals.md
@@ -6,7 +6,7 @@ By default, Honeycomb includes events for all web vitals except first-input-dela
 
 All metrics have the following attributes: `web_vital.name`, `web_vital.rating`, `web_vital.id`
 
-```
+```ts
 /**
  * The name of the metric (in acronym form).
  */
@@ -29,12 +29,11 @@ All metrics have the following attributes: `web_vital.name`, `web_vital.rating`,
  */
 
 "web_vital.id": string;
-
 ```
 
 ### [Cumulative Layout Score](https://web.dev/articles/cls) attributes
 
-```
+```ts
 /* Shared fields */
 "web_vital.name": "CLS",
 "web_vital.rating": "good" | "needs-improvement" | "poor",
@@ -45,6 +44,7 @@ All metrics have the following attributes: `web_vital.name`, `web_vital.rating`,
 "cls.value": number;
 "cls.delta": number;
 
+ /**
  * A selector identifying the first element (in document order) that
  * shifted when the single largest layout shift contributing to the page's
  * CLS score occurred.
@@ -86,7 +86,7 @@ The event time is when the single largest layout shift contributing to the page'
 
 ### [Largest Contentful Paint](https://web.dev/articles/lcp) attributes
 
-```
+```ts
 /** Shared fields */
 "web_vital.name": "LCP",
 "web_vital.rating": "good" | "needs-improvement" | "poor",
@@ -109,7 +109,7 @@ The event time is when the single largest layout shift contributing to the page'
  */
 "lcp.url": string;
 
-**
+/**
  * The time from when the user initiates loading the page until when the
  * browser receives the first byte of the response (a.k.a. TTFB). See
  * [Optimize LCP](https://web.dev/articles/optimize-lcp) for details.
@@ -143,7 +143,7 @@ The event time is the start of the page load (performance.timeOrigin), the durat
 
 ### [Interaction to Next Paint](https://web.dev/articles/optimize-inp) attributes
 
-```
+```ts
 /** Shared fields */
 "web_vital.name": "INP",
 "web_vital.rating": "good" | "needs-improvement" | "poor",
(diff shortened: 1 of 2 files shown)


docs/web-vitals.md
```diff
@@ -0,0 +1,362 @@
...
+  onFCP?: FCPReportCallbackWithAttribution;
+  onFID?: FIDReportCallbackWithAttribution;
+  onINP?: INPReportCallbackWithAttribution;
+  onLCP?: LCPReportCallbackWithAttribution;
+  onTTFB?: TTFBReportCallbackWithAttribution;
+
+  /** Whether or not the metric's value should be sent as
+   * `web_vital.value` as well as `[metricName].value`, ie `lcp.value`.
+   * Defaults to false */
+  includeValueInTopLevelNamespace?: boolean;
```
ahrbnsn, line 359: `includeValueInTopLevelNamespace` is a pretty wretched name; not proposing as the actual name if we go with this config; i just couldn't for the life of me come up with anything else & thought it was _so_ bad it couldn't be ignored & would definitely get a replacement

pkanal:
Thanks for putting so much thought into all of this decision making! In general I am on board with removing the `web_vital` namespace. Looking at the [OTel Semantic Conventions](https://github.com/scheler/opentelemetry-specification/blob/browser-events/specification/logs/events/semantic_conventions/browser.md#webvital) for web vitals, it looks like they don't have any name space to begin with. Instead the structure is that the event name would be `web_vital` and then the `name` field would be the specific web vital. This poses a problem because at the moment, `name` corresponds to the span name. Generally, I would prefer to stick as close to the OTel semantic conventions as possible to allow for smoother integration when this instrumentation becomes available upstream.

I think the best way to do that would be:
- We drop all attribute namespacing
- We have a `type` attribute that gets set to `web_vital` for all web vitals spans

Lemme know what you think!

docs/web-vitals.md
```diff
@@ -0,0 +1,362 @@
...
+## Additional Configuration
+
+### Using custom data attribute to identify elements
+
+Sometimes the element selector (used by LCP, INP & CLS events) is not very human-readable (or changes frequently) & makes it difficult to figure out the actual culprit. Pass in a custom data selector & hny will use that identifier instead for the `lcp.element`, `cls.element` & `inp.element` attributes.
+
+```js
+const sdk = new HoneycombWebSDK({
+  serviceName: 'my-app',
+  webVitals: {
```
pkanal, line 296: Maybe this API should be `webVitalsInstrumentation`? And this is where we can also enable/disable it? The other thought I had was keeping it consistent with how other auto-instrumentation is configured which would be:

```js
  instrumentations: [new WebVitalsInstrumentation({
    enabled: true,
    elementDataAttribute: 'human-id'
  })]
```

The upside of this is keeping it consistent with other auto-instrumentation and not having to maintain another layer between our API and the instrumentation. The downside is having to import and inititalize `new WebVitalsInstrumentation({})` if the user wants any customization or to disable the instrumentation.

JamieDanielson: 👍  I like the adjustment of `webVitalsInstrumentation`.

🤔  I also like the second suggestion that makes it closer to upstream auto-instrumentation setup... but acknowledge the potential yuck factor with requiring someone to do that extra work to customize. If this instrumentation does get added upstream though, it's likely to have this type of interface. So is it better to go with that now? 

Also I guess I'm curious about how much data this generates and maybe we need a PM input on how common this will need to be adjusted, and if you're at the point where you want to make those customizations, then step-by-step instructions regarding adding another package maybe isn't terrible?

docs/web-vitals.md
```diff
@@ -0,0 +1,362 @@
+# Web Vitals
+
+By default, Honeycomb includes events for all web vitals except first-input-delay. First-input-delay is set to be replaced by interaction-to-next-paint as a core web vital in March 2024.
```
JamieDanielson, line 3: Can we add a link here for reference? Ideally it won't be long-lived but if the date moves and/or folks are still used to that first-input-delay, we can avoid confusion a bit.

docs/web-vitals.md
```diff
@@ -0,0 +1,362 @@
...
+
+## Metric Attributes
```
JamieDanielson, line 5: This might sound pedantic, but I don't _think_ it is, or at least I don't intend it to be. It's because of an unfortunate collision between existing naming/usage and our desired usage.... Anywhere we can avoid using the term "metric" and replace it with "vital" I think that would helpful, to distinguish that we are not using the metric signal here.

ahrbnsn: no, i love it — definitely not pedantic, +++ clarity

docs/web-vitals.md
ahrbnsn: i think for me it hinges on whether or not we want to ship web-vital support out-of-the-box, by default. 

if yes, i'm inclined towards 

```js
  serviceName: 'my-app',
  webVitalsInstrumentation: {//...custom config},
})

```

When the config is absent, we rig our default set-up. if false, opt out. 

but if we don't want a "web-vitals, batteries-included" version of the wrapper, 
```
instrumentations: [new WebVitalsInstrumentation({
    enabled: true,
    elementDataAttribute: 'human-id'
  })]
```

definitely makes more sense (and we could probably scrape the `enabled` option altogether)

src/types.ts
```diff
@@ -95,4 +95,21 @@ export interface HoneycombOptions extends Partial<WebSDKConfiguration> {
...
+  reportOptions?: {
+    reportAllChanges?: boolean;
+    durationThreshold?: number;
+  };
+  // onCLS?: CLSReportCallbackWithAttribution;
+  // onFCP?: FCPReportCallbackWithAttribution;
+  // onFID?: FIDReportCallbackWithAttribution;
+  // onINP?: INPReportCallbackWithAttribution;
+  // onLCP?: LCPReportCallbackWithAttribution;
+  // onTTFB?: TTFBReportCallbackWithAttribution;
```
ahrbnsn, line 114: commented out for now— these are all types we'd import from the `web-vitals/attribution` package

ahrbnsn:
> Thanks for putting so much thought into all of this decision making! In general I am on board with removing the `web_vital` namespace. Looking at the [OTel Semantic Conventions](https://github.com/scheler/opentelemetry-specification/blob/browser-events/specification/logs/events/semantic_conventions/browser.md#webvital) for web vitals, it looks like they don't have any name space to begin with. Instead the structure is that the event name would be `web_vital` and then the `name` field would be the specific web vital. This poses a problem because at the moment, `name` corresponds to the span name. Generally, I would prefer to stick as close to the OTel semantic conventions as possible to allow for smoother integration when this instrumentation becomes available upstream.
> 
> I think the best way to do that would be:
> 
>     * We drop all attribute namespacing
> 
>     * We have a `type` attribute that gets set to `web_vital` for all web vitals spans
> 
> 
> Lemme know what you think!

mainly due to my eager desire to blunt some of honeycomb's (the product/ui) rough edges, i've become a pretty big fan of namespacing. i found it really helps w/ discoverability & successful querying.

once a dataset gets decently robust--a thriving & diverse ecosystem of events--figuring out what fields to query for a particular event type can become a truly terrible game of taking one's best guess--a wild stab in the dark--& frequently coming up empty handed, trying again, no dice, & eventually giving up and just going to the events tab. Watching spinners spin & twirling one's thumbes while hny fetches a list of available fields. not zippy or fun. i think it's also notable the "not knowing what i can query" affliction most acutely hits folks less familiar w/ the dataset (or even honeycomb overall), which adds extra drag, demotivating, to the folks who would most benefit from an easier on-ramp.

the more event types, the longer the list of query-able fields & the lower the likelihood a specific attribute exists on a specific span type. without namespacing, we need to rely on memory or code-spelunking to figure out what attributes are viable (or load up the events tab). which can be a bummer, or at least add friction.

with namespaced attributes, we get much higher discoverability — "this is a field on this specific event type, might be worth a look!". & can act as breadcrumbs for the curious who just want to explore their data & poke around.

Screenshot 2024-02-13 at 7 54 05 AM

(let me know if i'm misunderstanding, i took "drop all attribute namespacing to mean `cls.value`/`web_vital.value` would become `value`)

w/ namespacing, we'd also gain the ability to (potentially) auto-add field descriptions to RUM datasets in the honeycomb UI, which feels like it could be a pretty low-effort way to add some sparkle (and context) for our users. ie - if we see a `cls.value` field coming from a service using the hny web distro, we can pre-fill out the field description with confidence. but we couldn't do that for a `value` field. 

also--riffing off what you mentioned re: the name attribute--the otel list of attribute names seems like it has a pretty high likelihood of collision, and i'd argue would benefit from namespacing. ie `id` seems like a prime candidate for a custom resource attribute & a plain `id` field strikes me as a (potential) foot-gun waiting to happen. 

& speaking from my own exeprience, it can be confusing as heck when fields get invisibly overwritten. in hny's current custom front-end telemetry, i once added an attribute `type` to a custom event, unaware that buried deep in our home-spun telemetry code, we were already using `type`  to set the event name - 
```js
// deep in the internal guts
const payload = { 
 type: event.name,
 ...event.payload
}
```
i couldn't figure out why i couldn't find my events until i eventually figured out it was because i had accidentally overwritten (what i thought of as) the event name when i passed in a `type` attribute.  (i'm perhaps overly-vigilante re: wanting to prevent other folks a similar experience)

last, after taking a more in-depth look at the semantic conventions doc (thank you for the link!), i want to be even more heretical & argue for another break from the otel conventions — the doc states "The events may be represented either as Span Events or Log Events". 

i think `INP`, `LCP`, `TTFB` & `FCP` are best represented as `span`s, with durations, and not `span events` (although `CLS` makes sense, to my mind, as a span event). The doc is listed as experimental (and not merged to main) - are there opportunities to advocate for upstream changes?

docs/web-vitals.md
ahrbnsn: 100%— added a link but also dropping a screenshot here for posterity 

Screenshot 2024-02-13 at 9 26 50 AM

pkanal:
As we chatted about as a team, I'm going to close this PR in favour of a new one!

PR #70, opened 2024-02-12
feat: Add DeterministicSampler for easier sampling

JamieDanielson:
## Which problem is this PR solving?

- Closes #67 

## Short description of the changes

- add Deterministic Sampler
- add sampleRate info to debug
- add unit tests and update smoke test

## How to verify that this has the expected result

```js
  const sdk = new HoneycombWebSDK({
    apiKey: HONEYCOMB_API_KEY,
    serviceName: 'web-distro',
    debug: true,
    sampleRate: 2,
    instrumentations: [getWebAutoInstrumentations()], // add auto-instrumentation
  });
```

With a sampleRate higher than one you may have to generate a few traces before it shows up in Honeycomb! But with debug enabled at least you can see the config logged, and also see the "Non recording span" if it's not going to send because it's being dropped with sampling.

PR #72, opened 2024-02-14
feat: add AlwaysOffSampler

JamieDanielson:

- #70 added deterministic sampling but converted a sample rate of 0 to 1, but it may be nice to have the option keep the AlwaysOffSampler as an option (we have this option in the node distro).

- Allow sample rate of 0 and don't convert to default of 1
- Add AlwaysOffSampler to be used with sample rate of 0; drop everything.
- Update tests

Setting a sample rate of 0 is valid and should drop all telemetry

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -72,7 +72,8 @@ export interface HoneycombOptions extends Partial<WebSDKConfiguration> {
   serviceName?: string;
 
   /** The sample rate used to determine whether a trace is exported.
-   * This must be a whole number greater than 1. Only 1 out of every `sampleRate` traces will be randomly selected to be sent.
+   * This must be a whole positive number. Only 1 out of every `sampleRate` traces will be randomly selected to be sent.
+   * Set to 0 to drop everything.
    * Defaults to 1 (send everything).
    */
   sampleRate?: number;
--- a/src/util.ts
+++ b/src/util.ts
@@ -72,10 +72,9 @@ export const getTracesApiKey = (options?: HoneycombOptions) => {
 export const getSampleRate = (options?: HoneycombOptions) => {
   if (
     // must be a whole positive integer
-    options?.sampleRate !== null &&
-    options?.sampleRate !== undefined &&
-    options?.sampleRate >= 0 &&
-    Number.isSafeInteger(options?.sampleRate) === true
+    typeof options?.sampleRate === 'number' &&
+    Number.isSafeInteger(options?.sampleRate) &&
+    options?.sampleRate >= 0
   ) {
     return options?.sampleRate;
   }
--- a/src/util.ts
+++ b/src/util.ts
@@ -71,10 +71,11 @@ export const getTracesApiKey = (options?: HoneycombOptions) => {
 
 export const getSampleRate = (options?: HoneycombOptions) => {
   if (
-    // sample rate must be a whole integer greater than 0
-    options?.sampleRate &&
-    options?.sampleRate > 0 &&
-    Number.isSafeInteger(options?.sampleRate)
+    // must be a whole positive integer
+    options?.sampleRate !== null &&
+    options?.sampleRate !== undefined &&
+    options?.sampleRate >= 0 &&
+    Number.isSafeInteger(options?.sampleRate) === true
   ) {
     return options?.sampleRate;
   }
--- a/src/deterministic-sampler.ts
+++ b/src/deterministic-sampler.ts
@@ -1,5 +1,6 @@
 import { Attributes, Context, Link, SpanKind } from '@opentelemetry/api';
 import {
+  AlwaysOffSampler,
   AlwaysOnSampler,
   Sampler,
   SamplingResult,
@@ -29,6 +30,10 @@ export class DeterministicSampler implements Sampler {
   constructor(sampleRate: number) {
     this._sampleRate = sampleRate;
     switch (sampleRate) {
+      // sample rate of 0 means send nothing
+      case 0:
+        this._sampler = new AlwaysOffSampler();
+        break;
       // sample rate of 1 is default, send everything
       case 1:
         this._sampler = new AlwaysOnSampler();
--- a/test/deterministic-sampler.test.ts
+++ b/test/deterministic-sampler.test.ts
@@ -65,22 +65,22 @@ describe('configureDeterministicSampler', () => {
     expect(result.attributes).toEqual({ SampleRate: 1 });
   });
 
-  test('sample rate of 0 configures inner AlwaysOnSampler', () => {
+  test('sample rate of 0 configures inner AlwaysOffSampler', () => {
     const options = {
       sampleRate: 0,
     };
     const sampler = configureDeterministicSampler(options);
     expect(sampler).toBeInstanceOf(DeterministicSampler);
-    expect(sampler.toString()).toBe('DeterministicSampler(AlwaysOnSampler)');
+    expect(sampler.toString()).toBe('DeterministicSampler(AlwaysOffSampler)');
 
     const result = getSamplingResult(sampler);
-    expect(result.decision).toBe(SamplingDecision.RECORD_AND_SAMPLED);
-    expect(result.attributes).toEqual({ SampleRate: 1 });
+    expect(result.decision).toBe(SamplingDecision.NOT_RECORD);
+    expect(result.attributes).toEqual({ SampleRate: 0 });
   });
 
   test('sample rate of -42 configures inner AlwaysOn Sampler', () => {
     const options = {
-      sampleRate: 0,
+      sampleRate: -42,
     };
     const sampler = configureDeterministicSampler(options);
     expect(sampler).toBeInstanceOf(DeterministicSampler);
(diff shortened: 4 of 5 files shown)


src/types.ts
```diff
@@ -72,7 +72,8 @@ export interface HoneycombOptions extends Partial<WebSDKConfiguration> {
   serviceName?: string;
 
   /** The sample rate used to determine whether a trace is exported.
-   * This must be a whole number greater than 1. Only 1 out of every `sampleRate` traces will be randomly selected to be sent.
+   * This must be a whole positive number. Only 1 out of every `sampleRate` traces will be randomly selected to be sent.
+   * Set to 0 to drop everything.
```
JamieDanielson, line 76: This might not be super common but it's an easy lever to pull so it's nice to have.

src/util.ts
```diff
@@ -71,10 +71,11 @@ export const getTracesApiKey = (options?: HoneycombOptions) => {
 
 export const getSampleRate = (options?: HoneycombOptions) => {
   if (
-    // sample rate must be a whole integer greater than 0
-    options?.sampleRate &&
-    options?.sampleRate > 0 &&
-    Number.isSafeInteger(options?.sampleRate)
+    // must be a whole positive integer
+    options?.sampleRate !== null &&
+    options?.sampleRate !== undefined &&
```
JamieDanielson, line 76: Falsy values in JS are wild. This is necessary as `options?.sampleRate` will be falsy and return the default sample rate (see [typescript playground here](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAIgUQGIEECqAZAKgfQMooCyAChgjgEopYIwC8MAjANwBQrAlmFAKYBOAMwCGwHjAAS4HgE9QAWwBGAeQAOUDuAgwA3qxj6YEIXJUAbHhSG8A-AC4YYAK6L+bAL7tQkWAHMeUPGMzCysxBgAKEDUNSDsJKVkQRVV1TQBKegA+HT0DDgEYcNyDAwB6Upg5R2gYBTEhGAB3AAsQcxgVEAgOdQA3MS5ePz5ikpgo1NiAOiMTc0teGABCOgYnU1MYADIt0ZKJmIhrGaD50OXVmEcwABMeAS4eG+3dsYMDzWPZ4IWxTIYAAwvPYGAByzjqfCmHAggQEPAAktweMNItFPic5iFeBlVgwoHxHDxRhldG99Hx-I4+GBxujpt8zrw2CUPCVKVBqbTEKhMLgCCQyJRqAh3GxWF4an4AqdsTwkEJTBBpPRCh9YvZJGAZPJlPSILjsmS8gUiuSYOVKtVYHUYA0Wm0xJ1un0BsjhiD9Oqjpifucdp66ZMfYy5TB-jAgQHzeCXFCYXDEe7+Gjg19Zb9cZcCUSSTlzRyuUHDumsb8WQY2QZCzT4Mh0Nh8ERSOQqDQxZ5NLBvQgTFBpJqErqUodVdo3GxJd39QDB9rEsl9WPDBnQvYgRPO5BHVNTCAfOEAAbSwJl0Kpw69tTSNL2AAk2hPq94F80V-7aTch4ylsYEs0O57gex7+Kefovt6AK3jAD5PmeEEzp+34WhUAL-tu5i7vuR5weB8qKsqr6QO+N73o+oHPvhSrSEREAkUhP4VH+kqAdhIEyvBVGEZB0GwRRnEKtRtFQV+jFMEAA))

ahrbnsn: `Number.isSafeInteger` should return false for null & undefined as well as non-integer numbers, so i think you could shorten this to:

```tsx
if (Number.isSafeInteger(options?.sampleRate) && options.sampleRate >= 0) {
 return options.sampleRate
}
```

src/types.ts
ahrbnsn: if we drop everything, do we still get logs if `debug` is true? like could someone set it to zero to see what events _would_ be sent but not actually send events over the wire?

ahrbnsn (approved):

src/types.ts
JamieDanielson: Ooh excellent question. I'll double check to confirm, but I'm pretty sure it _won't_ log the events because the debug logger is intended to show events that will be sent, and this stops them from getting that far. This is more intended as an emergency lever, or closer to a no-op, to quickly disable further processing and exporting of telemetry.

src/util.ts
JamieDanielson: I guess it may be a TypeScript thing, but I can't get that second part of the equation in there, because I get an error of `'options.sampleRate' is possibly 'undefined'.`

ahrbnsn: (if the `Number.isSafeInteger(options?.sampleRate)` check passes, typescript knows options.sampleRate both exists & is a number, so you don't need the conditional chaining -- `options?.sampleRate` -- for any thing that follows)

JamieDanielson: This is what you mean right? I could definitely be missing something still, but I think TypeScript is not happy with this.

sampleRate could be undefined

JamieDanielson: Okay! Yeah it seems like that's how it should work for exactly why you mention it. But it must be something about the static analysis of the code or something.

ahrbnsn: ah... i'm used to writing in a codebase w/out `strictNullChecks` enabled: 

[sloppy ts setting](https://www.typescriptlang.org/play?strict=false&strictNullChecks=false&alwaysStrict=false#code/MYewdgzgLgBAIgUQGIEECqAZAKgfQMooCyAChgjgEopYIwC8MAjANwBQrAlmFAKYBOAMwCGwHjAAS4HgE9QAWwBGAeQAOUDuAgwA3qxj6YEIXJUAbHhSG8A-AC4YYAK6L+bAL7tQkWAHMeUPGMzCysxBgAKEDUNSDsJKVkQRVV1TQBKegA+HT0DDgEYcNyDAwB6Upg5R2gYBTEhGAB3AAsQcxgVEAgOdQA3MS5ePz5ikpgo1NiAOiMTc0teGABCOgYnU1MYADIt0ZKJmIhrGaD50OXVmEcwABMeAS4eG+3dsYMDzWPZ4IWxTIYAAwvPYGAByzjqfCmHAggQEPAAktweMNItFPic5iFeBlVgwoHxHDxRhldG99Hx-I4+GBxujpt8zrw2CUPCVKVBqbTEKhMLgCCQyJRqAh3GxWF4an4AqdsTwkEJTBBpPRCh9YvZJGAZPJlPSILjsmS8gUiuTwS4oTC4Yjkaj1UdMT9QhkdnTJhAnUy-oCSTlyTAOVz3YcvXKWQY2QYgzT4Mh0Nh8ERSOQqDQxZ5NLAHQgTFBpJqErqUodVdo3GxJdn9QDC9rEsl9WXDLLfvYgRXM5A2jwpqYQD5wgADaWBLG-NEe3NqaRpewAEm0o9boUnh2n+bSbiHGXKTAlmh7fYHw+X49XDoBc5gi7Pzt4a80V+3u4qAIP3fMx8HI-8Y-v8qKsqj6QBus4Lkuf4rrwCpKtIIEQGBW47jAe6MB+EBHv2P53t6sHAZe163lB54wUB8GES+qEVIwQA)

ahrbnsn: sorry for the red herring!

src/types.ts
JamieDanielson: 🤔  Something to consider though... I suppose we might like a ConsoleSpanExporter easy-button. Since debug _only_ logs if something is successfully exported, a "dev" option could ignore everything else and just always send to stdout and never try to export elsewhere, for exactly what you suggest here. I'll make an issue for consideration.

src/util.ts
JamieDanielson: 😄  "sloppy ts setting". No worries, if we decide we want that setting set to false later we can totally do that and simplify this code. I don't have super strong feels, though tend to lean toward stricter type checking to save me from myself (also lazy, this one is a default). But this is one of those scenarios that definitely gets people to shake their fist at TypeScript so I totally get it ha.

ahrbnsn: i don't know if this is any better, but adding a proper typecheck got strict typescript's begrudging acceptance:

```ts
const getSampleRateFalsy = (options?: HoneycombOptions) => {
    if (typeof options?.sampleRate !== "number" || !Number.isSafeInteger(options.sampleRate) || options.sampleRate < 0) {
        return DEFAULT_SAMPLE_RATE;
    }

    return options.sampleRate;
};
```

src/types.ts
JamieDanielson: - https://github.com/honeycombio/honeycomb-opentelemetry-web/issues/73

src/util.ts
JamieDanielson: I'm leaning toward this actually, also because I like that early return style too (you've mentioned in another PR, and I had something like this for a minute before switching back).

JamieDanielson: I changed my mind again on the early return only because I wasn't a fan of all the negations... BUT still took your suggestion to use the typeof check! Thanks for going on this journey with me! https://github.com/honeycombio/honeycomb-opentelemetry-web/pull/72/commits/d1e235e0dd9d35db66c24012b65ee1d351222649

PR #74, opened 2024-02-14
feat: add BaggageSpanProcessor

### LINKED ISSUE #66: [feat] Add BaggageSpanProcessor to allow multi-span attributes

Add a BaggageSpanProcessor to use in the web SDK, similar to what exists today in the Node Distro, to easily add attributes to downstream spans in a trace using baggage.

#13 specifically references using a shared package extracted out from the Node Distro and packaged separately to use in both places. For now we're going to essentially copy the functionality into here, and separate the shared dependency later.

JamieDanielson:

- Closes #66 

- Add BaggageSpanProcessor, and configure it in the span processor builder
- Add and update tests to confirm baggage gets added as span attributes
- Add debug log for entries in baggage

In the example app, add something to baggage and see it show up as an attribute on a span created from parent context:

```js
  const tracer = trace.getTracer('example-tracer');
  // new context based on current, with key/values added to baggage
  const ctx = propagation.setBaggage(
    context.active(),
    propagation.createBaggage({
      for_the_spans: { value: 'important value' },
    }),
  );
  // starting a new span on the tracer using the saved context
  context.with(ctx, () => {
    tracer.startActiveSpan('heygirl', (span) => {
      // do work here
      span.end();
    })
  })
```

baggage is span attribute

or nest even further to get on downstream spans as well

```js
  const tracer = trace.getTracer('example-tracer');
  // new context based on current, with key/values added to baggage
  const ctx = propagation.setBaggage(
    context.active(),
    propagation.createBaggage({
      for_the_spans: { value: 'important value' },
    }),
  );
  // starting a new span on the tracer using the saved context
  context.with(ctx, () => {
    tracer.startActiveSpan('heygirl', (span) => {
      // do work here
      context.with(trace.setSpan(context.active(), span), () => {
        tracer.startActiveSpan('heygirl2', (childspan) => {
          // do work here
          childspan.end();
        });
        span.end();
      })
    })
  });
```

image

### DIFF
--- a/src/baggage-span-processor.ts
+++ b/src/baggage-span-processor.ts
@@ -1,0 +1,43 @@
+import { Context, propagation, Span } from '@opentelemetry/api';
+import { SpanProcessor } from '@opentelemetry/sdk-trace-base';
+
+/**
+ * A {@link SpanProcessor} that reads entries stored in {@link Baggage}
+ * from the parent context and adds the baggage entries' keys and values
+ * to the span as attributes on span start.
+ *
+ * Keys and values added to Baggage will appear on subsequent child
+ * spans for a trace within this service *and* be propagated to external
+ * services in accordance with any configured propagation formats
+ * configured. If the external services also have a Baggage span
+ * processor, the keys and values will appear in those child spans as
+ * well.
+ *
+ * ⚠ Warning ⚠️
+ *
+ * Do not put sensitive information in Baggage.
+ *
+ * To repeat: a consequence of adding data to Baggage is that the keys and
+ * values will appear in all outgoing HTTP headers from the application.
+ */
+export class BaggageSpanProcessor implements SpanProcessor {
+  constructor() {}
+
+  onStart(span: Span, parentContext: Context): void {
+    (propagation.getBaggage(parentContext)?.getAllEntries() ?? []).forEach(
+      (entry) => {
+        span.setAttribute(entry[0], entry[1].value);
+      },
+    );
+  }
+
+  onEnd() {}
+
+  forceFlush() {
+    return Promise.resolve();
+  }
+
+  shutdown() {
+    return Promise.resolve();
+  }
+}
--- a/src/baggage-span-processor.ts
+++ b/src/baggage-span-processor.ts
@@ -1,4 +1,4 @@
-import { Context, propagation, Span } from '@opentelemetry/api';
+import { Context, diag, propagation, Span } from '@opentelemetry/api';
 import { SpanProcessor } from '@opentelemetry/sdk-trace-base';
 
 /**
@@ -27,6 +27,9 @@ export class BaggageSpanProcessor implements SpanProcessor {
     (propagation.getBaggage(parentContext)?.getAllEntries() ?? []).forEach(
       (entry) => {
         span.setAttribute(entry[0], entry[1].value);
+        diag.debug(
+          `@honeycombio/opentelemetry-web: 🚨 Baggage in all outgoing headers: ${entry[0]}=${entry[1].value} `,
+        );
       },
     );
   }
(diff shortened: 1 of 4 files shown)


src/span-processor-builder.ts
```diff
@@ -18,12 +19,15 @@ import { configureHoneycombHttpJsonTraceExporter } from './http-json-trace-expor
...
   const honeycombSpanProcessor = new CompositeSpanProcessor();
 
-  // We have to configure the exporter here becuase the way the base SDK is setup
+  // We have to configure the exporter here because the way the base SDK is setup
   // does not allow having both a `spanProcessor` and `traceExporter` configured.
   honeycombSpanProcessor.addProcessor(
     new BatchSpanProcessor(configureHoneycombHttpJsonTraceExporter(options)),
   );
 
+  // we always want to add the baggage span processor
```
JamieDanielson, line 28: consider for later: make this configurable? It's not configurable in the other distros but we've talked about it in case someone doesn't want it.

src/baggage-span-processor.ts
```diff
@@ -0,0 +1,43 @@
...
+ * to the span as attributes on span start.
+ *
+ * Keys and values added to Baggage will appear on subsequent child
+ * spans for a trace within this service *and* be propagated to external
+ * services in accordance with any configured propagation formats
+ * configured. If the external services also have a Baggage span
+ * processor, the keys and values will appear in those child spans as
+ * well.
+ *
+ * ⚠ Warning ⚠️
```
JamieDanielson, line 16: This whole warning came from the wording we had in the node distro. Is it too intense or is it just right? 🤷

JamieDanielson:
I think a solid follow-up PR is going to be to have a more detailed example app, whether that's the main example or a separate example, to show this usage along with some of the other things like the resource niceties Ashley added - that will allow us to easily smoke test those things as well.

test/baggage-span-processor.test.ts
```diff
@@ -0,0 +1,45 @@
+import { BaggageSpanProcessor } from '../src/baggage-span-processor';
```
JamieDanielson, line 1: While the setup of the processor itself was different in the node distro, this specific test is pretty much copy-pasta but it seemed to still do what we need

src/baggage-span-processor.ts
ahrbnsn: I'd say you could go even _more_ intense if the mood suited you. Emoji explosion. 

and/or maybe we could add some spicy logs when debug mode is active?

```🚨🚨🚨🚨🚨🚨 ADDED xxxx TO BAGGAGE 🚨🚨🚨🚨🚨🚨```
```BROADCASTING PLAIN VALUE IN ALL HEADERS```

i love that you're calling it out

File: src/span-processor-builder.ts (hunk 1 above)
Comment (ahrbnsn) [reply src/span-processor-builder.ts:28]: do you think it'd be useful to add the session id to the baggage? thinking it would help link server-side traces to specific sessions

File: test/span-processor-builder.test.ts [hunk 4]
```diff
@@ -97,15 +102,19 @@ describe('configureSpanProcessors', () => {
...
     );
-
-    honeycombSpanProcessors.onStart(span, ROOT_CONTEXT);
+    const bag = propagation.createBaggage({
+      'app.message': { value: 'heygirl' },
+    });
+    const ctx = propagation.setBaggage(ROOT_CONTEXT, bag);
+    honeycombSpanProcessors.onStart(span, ctx);
     expect(span.attributes).toEqual({
+      'app.message': 'heygirl',
```
ahrbnsn, line 117: ❤️ ❤️ ❤️

ahrbnsn (approved):
🧨 :

src/span-processor-builder.ts
JamieDanielson: Ooh adding session id sounds like a great idea! Will set a follow-up for that.

src/baggage-span-processor.ts
JamieDanielson: Now if debug is enabled this shows up in the console:
baggage in all outgoing headers

ahrbnsn: 😻 😻 😻

src/span-processor-builder.ts
JamieDanielson: - #75

PR #77, opened 2024-02-15
feat: ✨ Add `skipOptionsValidation` option and validate options

### LINKED ISSUE #68: [feat] Add skipOptionsValidation or similar for silencing warnings

When sending to an OpenTelemetry Collector, you don't want the apiKey in the config because it's in the collector config. However, we emit warnings when an API Key is missing from the config.

We can add the `skipOptionsValidation` config option to silence these warnings, similar to what we have in the node distro.

Maybe a better name or configuration for this? 

**Additional context**

This is listed already in our SDK Configuration and has some parts built but commented out in the repo already.

pkanal:

- Closes #68 

Adds validation warnings for options. 
Adds `skipOptionsValidation` option to config. This option is used when sending directly to a collector so warnings for not providing an `apiKey` can be suppressed.

- Warns if there is an `apiKey` missing
- Warns if there is a `serviceName` missing
- Warns if a `dataset` is provided but the `apiKey` is a E&S key and a `serviceName is provided
- Warns if `dataset` is missing if a Classic `apiKey` is provided
- Warns if `sampler` has been overridden
- Adds `skipOptionsValidation` option that doesn't show any of these warnings  if set to `true`, set to `false` by default

- Run the example app and don't provide an `apiKey`, you should see the `apiKey` warning and so forth
- Run the example app and set `skipOptionsValidation` to `true`, you should see the skip options validation message and no other warnings
- Tests pass

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -18,9 +18,9 @@ export const defaultOptions: HoneycombOptions = {
   serviceName: DEFAULT_SERVICE_NAME,
   debug: false,
   sampleRate: 1,
+  skipOptionsValidation: false,
   // TODO: Not yet implemented
   // localVisualizations: false,
-  // skipOptionsValidation: false,
 };
 
 export const MISSING_API_KEY_ERROR = `❌ @honeycombio/opentelemetry-web: Missing API Key. Set \`apiKey\` in HoneycombOptions. Telemetry will not be exported.`;
--- a/src/util.ts
+++ b/src/util.ts
@@ -23,8 +23,8 @@ export const defaultOptions: HoneycombOptions = {
   // localVisualizations: false,
 };
 
-export const MISSING_API_KEY_ERROR = `❌ @honeycombio/opentelemetry-web: Missing API Key. Set \`apiKey\` in HoneycombOptions. Telemetry will not be exported.`;
-export const MISSING_SERVICE_NAME_ERROR = `❌ @honeycombio/opentelemetry-web: Missing Service Name. Set \`serviceName\` in HoneycombOptions. Defaulting to '${defaultOptions.serviceName}'`;
+export const createHoneycombSDKLogMessage = (message: string) =>
+  `@honeycombio/opentelemetry-web: ${message}`;
 
 /**
  * Determines whether the passed in apikey is classic (32 chars) or not.
--- a/src/validate-options.ts
+++ b/src/validate-options.ts
@@ -41,7 +41,7 @@ export const validateOptionsWarnings = (options?: HoneycombOptions) => {
 
   // warn if custom sampler provided
   if (options?.sampler) {
-    console.warn(SAMPLER_OVERRIDE_WARNING);
+    console.debug(SAMPLER_OVERRIDE_WARNING);
   }
 
   return options;
--- a/src/validate-options.ts
+++ b/src/validate-options.ts
@@ -6,15 +6,15 @@ import {
 } from './util';
 
 export const IGNORED_DATASET_ERROR =
-  'WARN: Dataset is ignored in favor of service name.';
+  '🔕 WARN: Dataset is ignored in favor of service name.';
 export const MISSING_DATASET_ERROR =
-  'WARN: Missing dataset. Specify either HONEYCOMB_DATASET environment variable or dataset in the options parameter.';
+  '❌ WARN: Missing dataset. Specify either HONEYCOMB_DATASET environment variable or dataset in the options parameter.';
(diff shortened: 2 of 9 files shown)


src/validate-options.ts
```diff
@@ -0,0 +1,48 @@
...
+  '🔕 WARN: Dataset is ignored in favor of service name.';
+export const MISSING_DATASET_ERROR =
+  '❌ WARN: Missing dataset. Specify either HONEYCOMB_DATASET environment variable or dataset in the options parameter.';
+export const SKIPPING_OPTIONS_VALIDATION_MSG =
+  '⏭️ DEBUG: Skipping options validation. To re-enable, set skipOptionsValidation option or HONEYCOMB_SKIP_OPTIONS_VALIDATION to false.';
+export const SAMPLER_OVERRIDE_WARNING =
+  '🔨 WARN: Default deterministic sampler has been overridden. Honeycomb requires a resource attribute called SampleRate to properly show weighted values. Non-deterministic sampleRate could lead to missing spans in Honeycomb. See our docs for more details. https://docs.honeycomb.io/getting-data-in/opentelemetry/node-distro/#sampling-without-the-honeycomb-sdk';
+
+export const validateOptionsWarnings = (options?: HoneycombOptions) => {
+  if (options?.skipOptionsValidation) {
```
pkanal, line 18: Decided to do the logic of skipping validation inside the function instead at the top level in the SDK because it meant we could keep the `SKIPPING_OPTIONS_VALIDATION_MSG` contained to this file

src/validate-options.ts
```diff
@@ -0,0 +1,48 @@
+import { HoneycombOptions } from './types';
+import {
+  isClassic,
+  MISSING_API_KEY_ERROR,
+  MISSING_SERVICE_NAME_ERROR,
+} from './util';
+
+export const IGNORED_DATASET_ERROR =
```
pkanal, line 8: I went back and forth on whether these should live in `util` or not. 🤷🏽‍♀️

test/honeycomb-debug.test.ts
```diff
@@ -24,12 +24,14 @@ describe('when debug is set to true', () => {
       new HoneycombWebSDK({
         debug: true,
       });
-      expect(consoleSpy.mock.calls[1][0]).toContain(
-        'Honeycomb Web SDK Debug Mode Enabled',
+      expect(consoleSpy).toHaveBeenNthCalledWith(
```
pkanal, line 27: Did a lil refactor here to use Jest functions instead of array lookups because I found it really hard to pinpoint the right message quickly with the arrays 🙈

JamieDanielson: Oh I love this!

JamieDanielson (commented):
This all works great, though I do feel like there is a lot of intertwining between this and the debug file. Both things are intended to help warn a dev about incorrect setup. I left a few notes, though not sure about the broader question about how best to merge or separate this functionality from debug. 

As an example, here is the output if debug mode is enabled, while setting an unnecessary dataset and missing a service name. Maybe some of that redundancy is fine though, especially with important missing items or if you want to skipValidation but still know in debug mode. What do you think?

consolewarn and debug

src/validate-options.ts
```diff
@@ -0,0 +1,48 @@
...
+  '🔕 WARN: Dataset is ignored in favor of service name.';
```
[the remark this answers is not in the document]
pkanal: Yes I so agree with this!

src/validate-options.ts
pkanal: Yeah it feels better to me to have them all in the same place and the `validate-options` file feels like the right place 👍🏽

src/util.ts
```diff
@@ -18,13 +18,13 @@ export const defaultOptions: HoneycombOptions = {
...
   sampleRate: 1,
+  skipOptionsValidation: false,
   // TODO: Not yet implemented
   // localVisualizations: false,
-  // skipOptionsValidation: false,
 };
 
-export const MISSING_API_KEY_ERROR = `❌ @honeycombio/opentelemetry-web: Missing API Key. Set \`apiKey\` in HoneycombOptions. Telemetry will not be exported.`;
-export const MISSING_SERVICE_NAME_ERROR = `❌ @honeycombio/opentelemetry-web: Missing Service Name. Set \`serviceName\` in HoneycombOptions. Defaulting to '${defaultOptions.serviceName}'`;
+export const createHoneycombSDKLogMessage = (message: string) =>
```
JamieDanielson, line 26: 🤩

JamieDanielson (approved):
skipit

PR #88, opened 2024-02-23
maint: add releaseyml for generated release notes

JamieDanielson:

- the `release.yml` will automatically sort PRs into groupings based on type label, making for less work needed to organize release notes

- add `release.yml`

repos that have this file and auto-generate release notes for releases will see items grouped under the appropriate header as long as the appropriate label is applied


ahrbnsn (approved):
🧨

PR #90, opened 2024-02-27
fix: update version and add version change to releasing doc

JamieDanielson:

- telemetry still shows 0.0.1 even though current release is 0.0.2. I missed updating this on the last release.

- add step to releasing doc to update the version
- update the version to 0.0.2 since that is the current
- add test checking that version in `version.ts` matches version in `package.json`

tests pass, we remember to update version next time 😅

### DIFF
--- a/test/version.test.ts
+++ b/test/version.test.ts
@@ -1,0 +1,9 @@
+import { VERSION } from '../src/version';
+
+describe('sdk version', () => {
+  it('matches package.json version', () => {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    const expectedVersion = require('../package.json').version;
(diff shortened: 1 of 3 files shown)


test/version.test.ts
```diff
@@ -0,0 +1,9 @@
+import { VERSION } from '../src/version';
+
+describe('sdk version', () => {
+  it('matches package.json version', () => {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    const expectedVersion = require('../package.json').version;
+    expect(VERSION).toBe(expectedVersion);
```
ahrbnsn, line 7: ❤️

ahrbnsn (approved):

PR #91, opened 2024-02-28
feat: ✨ Config for web vitals

pkanal:
Adds extra config options for web vitals.

- Support for adding extra attributes to web vital spans
- Support for web-vitals options (e.g. reportAllChanges)

- Tests pass
- Test options out with the example app.


JamieDanielson (approved):
This is great! Let's set a follow-up for documenting this with JSDocs, and maybe adding to the example app that gets smoke tested 👍

PR #101, opened 2024-03-04
feat: add support for classic ingest keys

cewkrupa:

We've released Ingest Keys, but need to update the key detection logic to allow Ingest Keys to be used to send data to Classic environments.

- updates the Classic API Key detection logic to understand the shape of a Classic Ingest Key
- updates tests with a more comprehensive list of cases

It should be possible to use [an ingest key](https://docs.honeycomb.io/working-with-your-data/settings/api-keys/#ingest-keys-1) pointed to a classic environment after this change. Please note that the ability to create ingest keys in classic environments isn't yet publicly available (since we need to update this library and others to understand classic ingest keys).

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -25,14 +25,25 @@ export const defaultOptions: HoneycombOptions = {
 export const createHoneycombSDKLogMessage = (message: string) =>
   `@honeycombio/opentelemetry-web: ${message}`;
 
+const classicKeyRegex = /^[a-f0-9]*$/;
+const ingestClassicKeyRegex = /^hc[a-z]ic_[a-z0-9]*$/;
+
 /**
- * Determines whether the passed in apikey is classic (32 chars) or not.
+ * Determines whether the passed in apikey is classic or not.
  *
  * @param apikey the apikey
  * @returns a boolean to indicate if the apikey was a classic key
  */
 export function isClassic(apikey?: string): boolean {
-  return apikey?.length === 32;
+  if (apikey === null || apikey === undefined || apikey.length === 0) {
+    return false;
+  }
+  else if(apikey.length === 32) {
+    return classicKeyRegex.test(apikey);
+  } else if(apikey.length === 64) {
+    return ingestClassicKeyRegex.test(apikey);
+  }
+  return false;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/util.ts
```diff
@@ -25,14 +25,25 @@ export const defaultOptions: HoneycombOptions = {
...
+ * Determines whether the passed in apikey is classic or not.
  *
  * @param apikey the apikey
  * @returns a boolean to indicate if the apikey was a classic key
  */
 export function isClassic(apikey?: string): boolean {
-  return apikey?.length === 32;
+  if (apikey === null || apikey === undefined || apikey.length === 0) {
+    return false;
+  }
```
cewkrupa, line 40: This behavior is slightly different from how we implemented the logic in [libhoney-js](https://github.com/honeycombio/libhoney-js/pull/406/files#diff-02ac5abd26d8410a1b28c76bb9e40a78226c3a991553b4bafeba363f81dd4a61R228-R230) -- an empty / undefined key will evaluate as _not_ classic, while in libhoney-js and other libhoneys it would evaluate to true. There were explicit tests for this already so I chose to match what was already here, but I'm open to making these match if that's desired.

src/util.ts
```diff
@@ -25,14 +25,25 @@ export const defaultOptions: HoneycombOptions = {
...
 /**
- * Determines whether the passed in apikey is classic (32 chars) or not.
+ * Determines whether the passed in apikey is classic or not.
  *
  * @param apikey the apikey
  * @returns a boolean to indicate if the apikey was a classic key
  */
 export function isClassic(apikey?: string): boolean {
-  return apikey?.length === 32;
+  if (apikey === null || apikey === undefined || apikey.length === 0) {
```
JamieDanielson, line 38:
```suggestion
  if (apikey == null || apikey.length === 0) {
```
I recently learned [one of the few scenarios](https://eslint.org/docs/latest/rules/eqeqeq#smart) to use `==` is when checking if something is null or undefined. A double-equals check against null works for null and undefined so we don't need to triple-equals check both

jharley (approved):

src/util.ts
JamieDanielson: Yeah I think it would've been a breaking change for libhoney otherwise

ahrbnsn (approved):
🚀

PR #103, opened 2024-03-07
docs: add auto-instrumentation details

pkanal:
Adds docs to the readme linking out to details about included auto-instrumentation


ahrbnsn (approved):
🎉  📔

PR #128, opened 2024-04-25
fix: add opentelemetry/core as a dependency to package.json

### LINKED ISSUE #127: `package.json` is missing `@opentelemetry/core` in `peerDependencies`
**Versions**
0.1.1

**Steps to reproduce**

1. In a project using `pnpm` and `webpack` install this package without adding any additional dependencies
2. Attempt to import this package and bundle it with webpack
3. Webpack should throw and an error for not finding `@opentelemetry/core`. This is because this package never specifies it in package.json and thus will never be resolved.

**Additional context**
`pnpm` and newer versions of `yarn` require maintainers to specify packages they don't bundle but are expected to have as peer dependencies. Otherwise, you are relying on the dependency manager to `hoist` the dependency which is never guaranteed. It's best to specify these packages as peer dependencies with major ranges unless you need a specific version

Aghassi:
Solves #127 

This will fix support for package managers like `pnpm` that require you to be very explicit about your dependencies

- Closes #127 

Abides by requirements outlined here https://pnpm.io/how-peers-are-resolved

Given this package doesn't produce a bundle and is distributed using transpiled source code you still need the dependency manager to have the user resolve what is imported. The source code of this project directly imports this package https://github.com/search?q=repo%3Ahoneycombio%2Fhoneycomb-opentelemetry-web%20%22%40opentelemetry%2Fcore%22&type=code, thus this project relies on its presence.

Import this project and try to run it through webpack using `pnpm` as the package manager for the project. It will fail because the package I'm adding in this diff is never resolved in node_modules

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@opentelemetry/api": "~1.8.0",
+        "@opentelemetry/core": "^1.24.0",
         "@opentelemetry/exporter-trace-otlp-http": "~0.50.0",
         "@opentelemetry/instrumentation": "~0.50.0",
         "@opentelemetry/opentelemetry-browser-detector": "~0.50.0",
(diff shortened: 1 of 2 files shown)


pkanal:
Thank you so much for your contribution! Since this is being imported directly by the source code, could you update it to be a direct dependency instead?

Aghassi:
@pkanal done

pkanal:
@Aghassi due to some setup issues on our side, forked PRs are not running the CircleCI workflow. I have just updated it to run but I think I will have to close / reopen this PR. Going to try that now.

MustafaHaddara:
Thank you so much @Aghassi! We had merged other PRs that introduced merge conflicts onto your PR and we wanted to get this merged quickly so we've gone ahead and resolved those for you in https://github.com/honeycombio/honeycomb-opentelemetry-web/pull/136

PR #129, opened 2024-04-25
feat: Add s/m/l screen size resource attribute

pkanal:
Adds a resource attribute that classifies screen sizes into `small`, `medium` and `large`.

- Adds `screen.size`: `small` | `medium` | `large`
- `small` is less than 768px
- `medium` is between 768px and 1024px
- `large` is greater than 1024px

All of these were based on common media query sizes.

Run the app and see the `screen.size` attribute added to every span.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ The SDK adds these fields to all telemetry:
 | `page.hostname` | custom | per-span |   | `docs.honeycomb.io`   |
 | `screen.width` | custom | static | Total available screen width in pixels.   | `780`   |
 | `screen.height` | custom | static |  Total available screen height in pixels | `1000`   |
-| `screen.size` | custom | static |  `small` (less than 768px), `medium` (769px - 1024px) or `large` (greater than 1024px), `unknown` if the size is missing. | `large`   |
+| `screen.size` | custom | static |  `small` (less than 768px), `medium` (769px - 1024px) or `large` (greater than 1024px), `unknown` if the size is missing. | 
 | `honeycomb.distro.version` | stable | static | package version | "1.2.3" |
 | `honeycomb.distro.runtime_version` | stable | static | | "browser" |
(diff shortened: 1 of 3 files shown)


src/browser-attributes-resource.ts
```diff
@@ -1,5 +1,15 @@
 import { Resource } from '@opentelemetry/resources';
 
+type ScreenSize = 'small' | 'medium' | 'large';
+
+export const computeScreenSize = (screenWidth: number): ScreenSize => {
+  if (screenWidth <= 768) return 'small';
+  else if (screenWidth > 768 && screenWidth <= 1024) return 'medium';
+  else if (screenWidth > 1024) return 'large';
+
+  return 'large';
```
wolfgangcodes, line 10: 🤔 We _should_ never get here, but maybe `unknown`  would allow us to detect/distinguish error/unexpected cases?

wolfgangcodes (approved):
Looks good! just a smol Q about if we'd like to future proof our ability to detect garbage values for `window.screen.width`. (Prob over cautious thinking on my part)

README.md
```diff
@@ -131,6 +131,7 @@ The SDK adds these fields to all telemetry:
 | `page.hostname` | custom | per-span |   | `docs.honeycomb.io`   |
 | `screen.width` | custom | static | Total available screen width in pixels.   | `780`   |
 | `screen.height` | custom | static |  Total available screen height in pixels | `1000`   |
+| `screen.size` | custom | static |  `small` (less than 768px), `medium` (769px - 1024px) or `large` (greater than 1024px), `unknown` if the size is missing. | `large`   |
```
MustafaHaddara, line 134: there's an extra column here

```suggestion
| `screen.size` | custom | static |  `small` (less than 768px), `medium` (769px - 1024px) or `large` (greater than 1024px), `unknown` if the size is missing. | 
```

MustafaHaddara (approved):

PR #130, opened 2024-04-26
feat: add browser.name attribute

MustafaHaddara:
Adds a `browser.name` attribute to events

Delegates to [ua-parser-js](https://www.npmjs.com/package/ua-parser-js) so the list of available browsers comes from there. At its core, this is just user-agent parsing, with all of the caveats that come with that: it's fragile, a best guess effort, etc. 

ex. I tested with [Arc](https://arc.net) and it's getting parsed as Chrome; this is expected since Arc doesn't change the User Agent at all. 

Run the app and see `browser.name` attached: 
image

### DIFF
--- a/src/browser-attributes-resource.ts
+++ b/src/browser-attributes-resource.ts
@@ -13,17 +13,24 @@ export const computeScreenSize = (screenWidth: number): ScreenSize => {
 
 export const computeBrowserName = (userAgent: string) => {
   const uaParser = new UAParser(userAgent);
-  return uaParser.getBrowser().name ?? 'unknown';
+  const { name, version } = uaParser.getBrowser();
+
+  return {
+    name: name ?? 'unknown',
+    version: version ?? 'unknown',
+  };
 };
 
 export function configureBrowserAttributesResource(): Resource {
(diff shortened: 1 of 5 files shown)


src/browser-attributes-resource.ts
```diff
@@ -10,13 +11,19 @@ export const computeScreenSize = (screenWidth: number): ScreenSize => {
...
+};
+
 export function configureBrowserAttributesResource(): Resource {
   return new Resource({
     'user_agent.original': navigator.userAgent,
     //https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent#mobile_tablet_or_desktop
     'browser.mobile': navigator.userAgent.includes('Mobi'),
     'browser.touch_screen_enabled': navigator.maxTouchPoints > 0,
     'browser.language': navigator.language,
+    'browser.name': computeBrowserName(navigator.userAgent),
```
pkanal, line 26: Wondering if we should also include the browser version since we have access to it?

MustafaHaddara: sure, yeah, I can do that

pkanal (approved):

PR #140, opened 2024-05-03
fix: Add ability to disable, don't clobber defaults when passing other config params.

wolfgangcodes:

When passing some config parameters, other defaults were not being honored. 
When `disable` 'd web vitals would still be reported.

Exposes a disable function and ensures that the call backs for web-vitals will not generate spans.
Destructuring defaults will now supply defaults as expects when supplying only some configuration options.

Configure instrumentation with `{enabled: false}` you should not see spans generated. 
Configure instrumentation with `{vitalsToTrack: ['FID'],}` you should see other defaults still being honored i.e. `{enabled: true}`

### DIFF
--- a/src/web-vitals-autoinstrumentation.ts
+++ b/src/web-vitals-autoinstrumentation.ts
@@ -112,10 +112,7 @@ export class WebVitalsInstrumentation extends InstrumentationAbstract {
     this.ttfbOpts = ttfb;
     this._setupCallbacks = false;
     this._isEnabled = enabled;
-
-    if (this._isEnabled) {
-      this.enable();
-    }
+    this._setupWebVitalsCallbacks();
   }
 
   init() {}
@@ -372,7 +369,6 @@ export class WebVitalsInstrumentation extends InstrumentationAbstract {
       return;
     }
     this._isEnabled = true;
-    this._setupWebVitalsCallbacks();
     this._diag.debug(`Instrumentation  enabled`);
     this._diag.debug(`Sending spans for ${this.vitalsToTrack.join(',')}`);
   }
--- a/src/web-vitals-autoinstrumentation.ts
+++ b/src/web-vitals-autoinstrumentation.ts
@@ -120,7 +120,7 @@ export class WebVitalsInstrumentation extends InstrumentationAbstract {
 
   init() {}
 
-  setupWebVitalsCallbacks() {
+  private _setupWebVitalsCallbacks() {
     if (this._setupCallbacks) return;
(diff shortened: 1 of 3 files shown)


src/web-vitals-autoinstrumentation.ts
```diff
@@ -20,12 +20,10 @@ import {
...
   TTFBMetricWithAttribution,
 } from 'web-vitals/attribution';
-import {
-  InstrumentationBase,
-  InstrumentationConfig,
-} from '@opentelemetry/instrumentation';
+import { InstrumentationConfig } from '@opentelemetry/instrumentation';
 import { Span } from '@opentelemetry/api';
 import { VERSION } from './version';
+import { InstrumentationAbstract } from '@opentelemetry/instrumentation/build/src/instrumentation';
```
ahrbnsn, line 26: because it's just a type, does it make any sense to copy-paste the lil dude & define it inline of this file instead of importing it from the nested directory structure? might be a cursed and toxic idea

ahrbnsn (approved):
🚀

pkanal (approved):
🎉

src/web-vitals-autoinstrumentation.ts
wolfgangcodes: @ahrbnsn addressed here 84008e9f0b7551b5938e84feeb7a06c5216fe1bd

src/web-vitals-autoinstrumentation.ts
```diff
@@ -45,6 +56,116 @@ interface VitalOpts extends ReportOpts {
   applyCustomAttributes: ApplyCustomAttributesFn;
 }
 
+// To avoid importing InstrumentationAbstract from:
+// import { InstrumentationAbstract } from '@opentelemetry/instrumentation/build/src/instrumentation';
+// When this is exposed we can import from there.
+export abstract class InstrumentationAbstract implements Instrumentation {
```
pkanal, line 62: My only $0.02 about importing this vs copying it in is that I'm worried about code drift from the base package but I'll leave it up to you! I definitely see advantages of doing it this way but wanted to mention it.

pkanal (approved):

src/web-vitals-autoinstrumentation.ts
wolfgangcodes: :100: We'll be sure to follow up if/when that base class is exposed, but in the meantime, we're at least inline with the what the library is publicly exposing which feels marginally safer.

PR #146, opened 2024-05-10
maint: adding more detail to the release guide

MustafaHaddara:
@wolfgangcodes and I were doing a release today and we spotted some spots where we could add some more detail


pkanal (approved):
🎉

PR #151, opened 2024-05-15
feat: drop axios requirement

MustafaHaddara:
Some users were unable to install this package due to issues with our transitive dependency on `axios`. 

It turns out, though, that we don't really need axios for much. This PR removes it.

Drop axios, switch to standard `fetch` API

- [x] unit tests pass
- [x] Install into test repo, confirm ConsoleTraceLinkExporter still does what we think it does

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,6 @@
         "@opentelemetry/sdk-trace-base": "~1.24.0",
         "@opentelemetry/sdk-trace-web": "~1.24.0",
         "@opentelemetry/semantic-conventions": "~1.24.0",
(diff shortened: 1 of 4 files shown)


pkanal (approved):
It doesn't have to be a part of this PR, but it would be nice to have smoke tests for this since this isn't covered by unit tests. Its lower priority since this is more of a debug tool but perhaps worth logging an issue for?

MustafaHaddara:
@pkanal I added some smoke tests, let me know if you want me to change anything else.

pkanal (approved):
Amazing ty!!! 🚢

PR #152, opened 2024-05-16
feat: Upgrade to web-vitals@4.0.0

wolfgangcodes:
Web vitals but released 4.0.0 which includes an updates to field names and additional for `LCPAttribution`, `INPAttribution` and `TTFBAttribution` and deprecated `onFID`. See the full changes [here](https://github.com/GoogleChrome/web-vitals/blob/v4.0.0/docs/upgrading-to-v4.md).

Upgrade to `web-vitals@4.0.0`. Full change log of the underlying changes can be found [here](https://github.com/GoogleChrome/web-vitals/blob/v4.0.0/docs/upgrading-to-v4.md).

For this release, we'll send both the updated field names and the deprecated field names, which will be removed in the next major release.
Additionally, `onFID` has been deprecated and will removed in the next major release.

For this instrumentation we have updated attribute payloads as follows:

### `LCP`:
- Updated `resource_load_time` -> `resource_load_duration`
- Deprecated `resource_load_time`

### `INP`:
- Updated `element` -> `interaction_target`
- Deprecated `element`
- Updated `event_type` -> `interaction_type`
- Deprecated `event_type`
- Added `interaction_time` 
- Added `input_delay`
- Added `interaction_target_element`
- Added `next_paint_time`
- Added `presentation_delay`
- Added `processing_duration`

### `TTFB`:
- Updated `waiting_time` -> `waiting_duration`
- Deprecated `waiting_time`
- Updated `dns_time` -> `dns_duration`
- Deprecated `dns_time`
- Updated `connection_time` -> `connection_duration`
- Deprecated `connection_time`
- Updated `request_time` -> `request_duration`
- Deprecated `request_time`
- Added `cache_duration`

See that the new new and updated properties are being reported.

### DIFF
--- a/src/web-vitals-autoinstrumentation.ts
+++ b/src/web-vitals-autoinstrumentation.ts
@@ -338,7 +338,7 @@ export class WebVitalsInstrumentation extends InstrumentationAbstract {
       url,
       timeToFirstByte,
       resourceLoadDelay,
-      resourceLoadTime,
+      resourceLoadDuration,
       elementRenderDelay,
     }: LCPAttribution = attribution;
     const attrPrefix = this.getAttrPrefix(name);
@@ -350,7 +350,8 @@ export class WebVitalsInstrumentation extends InstrumentationAbstract {
       [`${attrPrefix}.url`]: url,
       [`${attrPrefix}.time_to_first_byte`]: timeToFirstByte,
       [`${attrPrefix}.resource_load_delay`]: resourceLoadDelay,
-      [`${attrPrefix}.resource_load_time`]: resourceLoadTime,
+      [`${attrPrefix}.resource_load_duration`]: resourceLoadDuration,
+      // [`${attrPrefix}.resource_load_time`]: resourceLoadDuration, // TODO:  Deprecate instead?
       [`${attrPrefix}.element_render_delay`]: elementRenderDelay,
     });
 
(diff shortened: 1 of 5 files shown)


src/web-vitals-autoinstrumentation.ts
```diff
@@ -448,21 +464,27 @@ export class WebVitalsInstrumentation extends InstrumentationAbstract {
...
-      [`${attrPrefix}.waiting_time`]: waitingTime,
-      [`${attrPrefix}.dns_time`]: dnsTime,
-      [`${attrPrefix}.connection_time`]: connectionTime,
-      [`${attrPrefix}.request_time`]: requestTime,
+      [`${attrPrefix}.waiting_duration`]: waitingDuration,
+      [`${attrPrefix}.dns_duration`]: dnsDuration,
+      [`${attrPrefix}.connection_duration`]: connectionDuration,
+      [`${attrPrefix}.request_duration`]: requestDuration,
+      [`${attrPrefix}.cache_duration`]: cacheDuration,
+      // [`${attrPrefix}.waiting_time`]: waitingDuration, // TODO:  Deprecate instead?
```
pkanal, line 484: I do like duration better than time in this case, BUT in order to not break any existing telemetry can we double send these values for a version while we make sure the dashboard is up to date with these values?

pkanal (approved):
🎉 Amazing! Can we these new attributes to our [docs for web vitals attributes](https://github.com/honeycombio/honeycomb-opentelemetry-web/blob/main/docs/web-vitals.md)? Up to you if you want to do it as a separate PR, that works too!

src/web-vitals-autoinstrumentation.ts
```diff
@@ -368,16 +370,33 @@ export class WebVitalsInstrumentation extends InstrumentationAbstract {
...
+      [`${attrPrefix}.input_delay`]: inputDelay,
+      [`${attrPrefix}.interaction_target`]: interactionTarget,
+      [`${attrPrefix}.interaction_time`]: interactionTime,
+      [`${attrPrefix}.interaction_type`]: interactionType,
       [`${attrPrefix}.load_state`]: loadState,
+      [`${attrPrefix}.next_paint_time`]: nextPaintTime,
+      [`${attrPrefix}.presentation_delay`]: presentationDelay,
+      [`${attrPrefix}.processing_duration`]: processingDuration,
+      // These will be deprecated in a future version
+      [`${attrPrefix}.element`]: interactionTarget,
```
ahrbnsn, line 398: definitely defer to purvi & other experts here, but i'd like to vote to keep `{attrPrefix}.element` as the designated field  (or at least dual write) — one struggle for folks less familiar with the specific data shape of various events is trying to remember what field name goes with what event type. ie instead of making people remember for `inp`, the element is listed under `interaction_target`, for `cls`, it's the `largest_shift_target`, and for `lcp`, it's `element`, keep all targets referenced under the same name — the element

more in-depth writeup/thoughts  here: https://github.com/honeycombio/honeycomb-opentelemetry-web/pull/61#issuecomment-1941634529

wolfgangcodes: That's great additional context! I think that makes me _more_ inclined to dual write. My _evolved thinking is that we should map 1-1 with the shape of web vitals data, and we can _additionally _ in a separate namespace provide more ergonomic access and/or derived data. Since we're still in EA and not removing any fields, let's merge this, and we open up a follow-up PR to tackle any changes as our thinking evolves.

PR #153, opened 2024-05-16
fix: Use constant for `service.name` from semconv library

pkanal:
While doing some bundle size analysis, noticed that the `@opentelemetry/semantic-conventions` library is a significant contributor to the bundle size because it has a lot of unminified constants. There was a [PR merged that also exports each constant by itself](https://github.com/open-telemetry/opentelemetry-js/pull/4298) instead of on an object which allows tree shaking to only import that constant instead of a whole object. This PR uses the new constants from the `@opentelemetry/semantic-conventions` instead of the deprecated `SemanticConventions` object.

- Use exported constant for service name.

- Smoke tests pass
- Service name is still respected, run the example app locally and check that service name is set to the value expected.

### DIFF
--- a/src/base-otel-sdk.ts
+++ b/src/base-otel-sdk.ts
@@ -38,7 +38,7 @@ import {
   WebTracerConfig,
   WebTracerProvider,
 } from '@opentelemetry/sdk-trace-web';
-import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions';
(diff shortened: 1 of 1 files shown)


MustafaHaddara (approved):
confirmed the new name lines up with the old name:

Screenshot 2024-05-16 at 12 30 57 PM
Screenshot 2024-05-16 at 12 31 10 PM

PR #181, opened 2024-06-21
add global errors instrumentation

nordfjord:

Sometimes errors escape and it's useful to know what those errors were.
OpenTelemetry suggests these should go on the log signal, which does
make sense, but honeycomb doesn't really use that signal so it makes
sense for this to live in the honeycomb distribution.

- Adds a new `GlobalErrorsInstrumentation`
- when enabled it attaches handlers to `unhandledrejection` and `error` events on the window
- The handler creates a span that matches the exception conventions from opentelemetry

- Run the tests

## Complications
- The jest environment is slightly different from actual browsers so I was unable to test the unhandled rejection flow, but have left a test that would work in an actual browser environment (marked as `it.skip` to avoid running it)
- Could not test with the instrumentation disabled because jest will fail any test that leaks an error automatically

### DIFF
--- a/src/global-errors-autoinstrumentation.ts
+++ b/src/global-errors-autoinstrumentation.ts
@@ -1,0 +1,52 @@
+import { InstrumentationConfig } from '@opentelemetry/instrumentation';
+import { InstrumentationAbstract } from './web-vitals-autoinstrumentation';
+import { VERSION } from './version';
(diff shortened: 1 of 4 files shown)


wolfgangcodes:
Closed with https://github.com/honeycombio/honeycomb-opentelemetry-web/pull/186 which includes these commits and a few other small tweaks.
Thanks again @nordfjord!

PR #189, opened 2024-07-08
feat: Update example to nest spans.

wolfgangcodes:
Without `ZoneContextManager` clicks that trigger network requests weren't nested. 

Using `ZoneContextManager`, we can see that spans are nested as expected
CleanShot 2024-07-08 at 17 49 47

Add `ZoneContextManager` to the hello world example so that spans nest.

Introspect debugger to see the the parent ids are set properly.


JamieDanielson (approved):
i love making connections

PR #191, opened 2024-07-11
feat: Add LOAF support for onINP

### LINKED ISSUE #138: Upgrade web-vitals.js library
There's a newer version of the web-vitals.js library that has much richer INP and LOAF data, can we explore using it?

**Additional context**

wolfgangcodes:

- Closes #138

Optionally add performance timing as spans for `INP` core web vital if `includeTimingsAsSpans` is set to true.

```js
webVitalsInstrumentationConfig: {
      vitalsToTrack: ['CLS', 'FCP', 'FID', 'INP', 'LCP', 'TTFB'],
      inp: {
        reportAllChanges: true,
        includeTimingsAsSpans: true,
      },
...
```

To force an poor INP you can add a snippet like this:
```js
const delay = (ms) => {
  const start = Date.now();
  while (Date.now() - start < ms) {
    // no-op
  }
};
```

And do:
```js
document.getElementById('my-button').addEventListener('click', () => {
    delay(1000);

    console.log(":(")
  });
```

if `includeTimingsAsSpans` is true you should see INP timing spans like this:
CleanShot 2024-07-31 at 13 06 39

### DIFF
--- a/packages/honeycomb-opentelemetry-web/examples/hello-world-web/index.js
+++ b/packages/honeycomb-opentelemetry-web/examples/hello-world-web/index.js
@@ -24,6 +24,9 @@ const main = () => {
       }),
     ],
     contextManager: new ZoneContextManager(),
+    webVitalsInstrumentationConfig: {
+      vitalsToTrack: ['CLS', 'FCP', 'FID', 'INP', 'LCP', 'TTFB'],
+    },
   });
   sdk.start();
   const tracer = trace.getTracer('click-tracer');
--- a/packages/honeycomb-opentelemetry-web/package-lock.json
+++ b/packages/honeycomb-opentelemetry-web/package-lock.json
@@ -21,7 +21,7 @@
         "@opentelemetry/semantic-conventions": "~1.25.1",
         "shimmer": "^1.2.1",
         "ua-parser-js": "^1.0.37",
-        "web-vitals": "^4.2.2"
+        "web-vitals": "^4.2.3"
       },
       "devDependencies": {
         "@babel/plugin-transform-runtime": "^7.24.0",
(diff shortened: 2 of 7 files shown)


src/web-vitals-autoinstrumentation.ts
```diff
@@ -379,31 +484,49 @@ export class WebVitalsInstrumentation extends InstrumentationAbstract {
...
+          [`${attrPrefix}.input_delay`]: inputDelay,
+          [`${attrPrefix}.interaction_target`]: interactionTarget,
+          [`${attrPrefix}.interaction_time`]: interactionTime,
+          [`${attrPrefix}.interaction_type`]: interactionType,
+          [`${attrPrefix}.load_state`]: loadState,
+          [`${attrPrefix}.next_paint_time`]: nextPaintTime,
+          [`${attrPrefix}.presentation_delay`]: presentationDelay,
+          [`${attrPrefix}.processing_duration`]: processingDuration,
+          [`${attrPrefix}.duration`]: inpDuration,
+          [`${attrPrefix}.timing.json`]: JSON.stringify(
```
pkanal, line 506: Is it possible for us to unpack the JSON object here instead of stringifying it? It makes querying this data a lot harder if it is JSON and more attributes are generally ok for us to add.

wolfgangcodes: Absolutely.

wolfgangcodes: >Absolutely.

This was a bit optimistic. Since this a JSON Array and not a JSON object it doesn't fit cleanly into `SpanAttributes` so I'm going to remove this for now and we can open a follow up PR with this data.

pkanal (approved):

PR #203, opened 2024-07-24
refactor: export multiple packages

MustafaHaddara:
This PR lets us publish multiple packages from this repo

This PR moves all of the main code to a `packages/honeycomb-opentelemetry-web/` folder, and sets up the barest scaffolding for a `packages/honeycomb-react-user-instrumentation/` package.

This also adjusts the existing CircleCI configuration to only run on changed files. Specifically, the main `.circleci/config.yml` now just determines which packages have changed, and then triggers the workflow(s) in `.circleci/build.yml` with the correct config options.

`.circle/build.yml` contains TWO workflows: one for the web distro and the other for the react instrumentation. Each workflow will only get triggered if files in that directory have changed. 

I also updated `RELEASING.md` with more details on our new release process. I am not sure how to test this without merging (is that even possible?) so I anticipate we may have to revisit those guidelines in the future, but for now reviewers should weigh in on whether we think the process described in that document is practical. 

I used [this pr](https://github.com/honeycombio/honeycomb-opentelemetry-web/pull/204) to test these changes, and the build/test pipeline there is currently green. 

This _should_ work for the release steps as well, but I have no way of testing that without merging this to main.


.circleci/config.yml
```diff
@@ -1,147 +1,21 @@
...
-          context: Honeycomb Secrets for Public Repos
-          requires:
-            - build
-            - smoke_test
+      - path-filtering/filter:
+          base-revision: main
+          config-path: .circleci/build-npm-package.yml
+          mapping: |
+            packages/honeycomb-opentelemetry-web/.* package-name "honeycomb-opentelemetry-web"
+            packages/honeycomb-react-user-instrumentation/.* package-name "honeycomb-react-user-instrumentation"
```
MustafaHaddara, line 21: from the group review: how difficult will it be for us to trigger builds for _all_ things that have changed

in the ideal world:
- dev pushes some changes
- circleci / npm / tooling stuff figures out what changed, and then:
  - runs tests for changed stuff
  - builds artifacts for changed stuff
  - creates a release for changed stuff

RELEASING.md
```diff
@@ -19,9 +24,13 @@
   - make sure the PR title starts with `rel:` to conform with our semantic commits convention
   - add the `no-changelog`, `type: maintenance`, and `version: no bump` labels
 - Once the pull request is approved, squash-merge it
+
+## Git Tags and Releases
 - Checkout `main` and fetch the now-updated `main` branch.
-- Apply a tag for the new version on the merged commit (e.g. `git tag -a v2.3.1 -m "v2.3.1"`)
-- Push the tag upstream (this will kick off the release pipeline in CI) e.g. `git push origin v2.3.1`
+- Apply a tag for the new version on the merged commit (e.g. `git tag -a v2.3.1 -m "web-distro-v2.3.1"`). The tag MUST begin with either `web-distro-v` or `react-instrumentation-v` in order to trigger our CI release pipeline
```
pkanal, line 30: should the -a also have the new prefix?

pkanal (commented):
I think it is possible to test the release process before merging to main by pushing up a dev tag like `web-distro-v2.3.1-dev-test` or something that would never be npm installed, we would just need to make sure it doesn't get tagged as the latest release in npm.

RELEASING.md
MustafaHaddara: whoops yeah i put it in the wrong spot 🙈

MustafaHaddara:
> I think it is possible to test the release process before merging to main by pushing up a dev tag like web-distro-v2.3.1-dev-test or something that would never be npm installed, we would just need to make sure it doesn't get tagged as the latest release in npm.

the problem is that when we push a tag, the circleci config on the `main` branch gets run...

I wonder if we could temporarily configure circleci to look at this branch instead?

pkanal:
> > I think it is possible to test the release process before merging to main by pushing up a dev tag like web-distro-v2.3.1-dev-test or something that would never be npm installed, we would just need to make sure it doesn't get tagged as the latest release in npm.
> 
> the problem is that when we push a tag, the circleci config on the `main` branch gets run...
> 
> I wonder if we could temporarily configure circleci to look at this branch instead?

Ah yes, I think we can change the circleci config in this PR temporarily to run for this branch and then it should work? And once we've confirmed it we can remove that from the config.

MustafaHaddara:
Wait, I'm silly-- I think I should be able to make a tag pointing _at this branch_ and then push that tag and watch circleci run. 🤦

MustafaHaddara:
Ok so I went ahead and pushed `web-distro-v0.0.0-test` ([tag here](https://github.com/honeycombio/honeycomb-opentelemetry-web/releases/tag/web-distro-v0.0.0-test))

That triggered this build https://app.circleci.com/pipelines/github/honeycombio/honeycomb-opentelemetry-web/590/workflows/0955903b-4424-4a2d-9d4d-ac642c344069

And resulted in this draft release https://github.com/honeycombio/honeycomb-opentelemetry-web/releases/tag/untagged-46a57f6c667eb538a24a

So I think it works!

pkanal (approved):
👏🏽 amazing, well done!!!

RELEASING.md
```diff
@@ -19,9 +24,13 @@
   - make sure the PR title starts with `rel:` to conform with our semantic commits convention
   - add the `no-changelog`, `type: maintenance`, and `version: no bump` labels
 - Once the pull request is approved, squash-merge it
+
+## Git Tags and Releases
 - Checkout `main` and fetch the now-updated `main` branch.
-- Apply a tag for the new version on the merged commit (e.g. `git tag -a v2.3.1 -m "v2.3.1"`)
-- Push the tag upstream (this will kick off the release pipeline in CI) e.g. `git push origin v2.3.1`
+- Apply a tag for the new version on the merged commit (e.g. `git tag -a web-distro-v2.3.1 -m "web-distro-v2.3.1"`). The tag MUST begin with either `web-distro-v` or `react-instrumentation-v` in order to trigger our CI release pipeline
```
wolfgangcodes, line 30: Does it make sense for our tag names (and build steps) to align more with package names rather than what we call Instead of  `web-distro `honeycomb-opentelemetry-web-vX.Y.Z` or `otel-web-vX.Y.Z` and `honeycomb-opentelemetry-react-user-instrumentation-vX.Y.Z` (or `react-user-instrumentation-vX.Y.Z`)  are more explicit especially as we add more packages.

RELEASING.md
```diff
@@ -19,9 +24,13 @@
...
+  - use the tag prefix that corresponds to the library you want to release
```
wolfgangcodes, line 31: Should we have a table of tags OR make a script package `npm run release-tag:all` or `npm run release-tag:otel-web-vX.Y.Z` for tagging and pushing? The script tag will document the tag names and make it less error prone since it's now more involved?

wolfgangcodes (approved):
Looks great! I would love to see our tags and build steps align with the packages they build and release but it's not a hill I'm going to die on, but I think it would help make things more clear to folks who weren't part of the this change and make the connections more clear for contributors.

RELEASING.md
MustafaHaddara: yeah that's fair. on one hand, I'm lazy and don't want to type a lot...but on the other, clarity is nice. I'll make that change.

RELEASING.md
MustafaHaddara: Automating more of this release process can be done in a later PR. For this PR, the important part is just getting to a multi-package world.

wolfgangcodes: Agree 💯 I'm lazy and don't want to typo a lot...so I'll followup with some scripts....eventually!

PR #239, opened 2024-08-12
chore: update gitignore paths

pkanal:
I noticed after running smoke tests locally that after the repo structure refactor, some files that should be ignored weren't being ignored.

- Updated gitignore


wolfgangcodes (approved):
🚀

PR #268, opened 2024-08-28
feat: Add Experimental User Interaction Instrumentation

MustafaHaddara:
supercedes https://github.com/honeycombio/honeycomb-opentelemetry-web/pull/243

Upstream's `UserInteractionInstrumentation` does not play well in React apps with many nested event handlers. This adds an experimental alternative `UserInteractionInstrumentation`.

Sets up an event listener at the top level instead of wrapping the global `addEventListener`. 

Smoke tests to come in a later PR. 

Run the example react app, click on some stuff. 

The app looks like this 
image

Clicks in the blue header should not create any spans.
Likewise, clicks below the outlined div should not create any spans.

Clicks anywhere in the middle div-- on the div itself, or on either button --should create spans.

You should be able to see the spans in the console log.


packages/experimental/test/user-interaction-instrumentation.test.ts
```diff
@@ -0,0 +1,22 @@
+import { wrapEventPropagationCb } from '../src/user-interaction-instrumentation';
+
+describe('wrapEventPropagationCb', () => {
+  it('wraps the function', () => {
```
wolfgangcodes, line 4:
```suggestion
  it('calls span.end on evt.stopPropagation', () => {
```

packages/experimental/src/index.ts
wolfgangcodes, line 1: I would expect the folder name to match the package name. I think that's a pretty common pattern. i.e `experimental` -> ` "experimental-opentelemetry-web"

MustafaHaddara: sure that's a simple change

packages/experimental/src/user-interaction-instrumentation.ts
```diff
@@ -0,0 +1,206 @@
...
+
+export class UserInteractionInstrumentation extends InstrumentationBase {
+  protected _config: UserInteractionInstrumentationConfig;
+  private _is_enabled: boolean;
+  private _listeners: Listener[];
+
+  constructor(config: UserInteractionInstrumentationConfig = {}) {
+    super(INSTRUMENTATION_NAME, VERSION, config);
+    this._config = config;
+    this._is_enabled = this._config.enabled ?? false;
```
wolfgangcodes: 🤔 usually we'd do `_isEnabled` but here we're doing 🐍 casing. Is that because of a collision with `InstrumentationBase`? I think I ran into a similar issue with `GlobalErrorInstrumentation`. I wasn't getting much value out off the base class and it wasn't reporting enabled correctly, so I just implemented `InstrumentationAbstract` instead.

MustafaHaddara: nah this might've just me been losing track, will update

packages/experimental/src/user-interaction-instrumentation.ts
```diff
@@ -0,0 +1,206 @@
...
+const shouldCreateSpan = (
+  event: Event,
+  element: EventTarget | null,
+  eventName: EventName,
+  rootNodeId: string | undefined,
+): element is HTMLElement => {
+  // @ts-expect-error we set this field in the handler, if it's already here then we've already seen this event
+  if (event[SPAN_KEY]) {
+    return false;
+  }
```
wolfgangcodes: I'd love to see this as a WeakMap or something so we can avoid modifying the event attributes, but this works for now!

packages/experimental/src/user-interaction-instrumentation.ts
```diff
@@ -0,0 +1,206 @@
...
+
+  if (!element.getAttribute) {
+    return false;
+  }
+
+  if (element.hasAttribute('disabled')) {
+    return false;
+  }
```
wolfgangcodes: 🤓 I was like surely, "Browsers don't emit events on disabled elements." Turns out there a lot of variability in the interpretation here. 😸

wolfgangcodes (approved):
Overall this looks like a great iteration!

packages/experimental/src/user-interaction-instrumentation.ts
MustafaHaddara: big sob indeed

PR #281, opened 2024-09-05
fix: Make applyCustomAttributes optional.

wolfgangcodes:

Makes the `ApplyCustomAttributesFn` optional since it's guarded against. 
Makes the `ApplyCustomAttributesFn` optional for `webVitalsInstrumentationConfig`, updates example app to exercise that.

### DIFF
--- a/packages/honeycomb-opentelemetry-web/src/web-vitals-autoinstrumentation.ts
+++ b/packages/honeycomb-opentelemetry-web/src/web-vitals-autoinstrumentation.ts
@@ -228,10 +228,6 @@ export interface WebVitalsInstrumentationConfig extends InstrumentationConfig {
   ttfb?: VitalOpts;
 }
 
-const noop = () => {
-  // No-op
-};
-
 /**
  * Web vitals auto-instrumentation, sends spans automatically for CLS, LCP, INP, FCP, FID, TTFB.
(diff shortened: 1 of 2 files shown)


packages/honeycomb-opentelemetry-web/src/web-vitals-autoinstrumentation.ts
```diff
@@ -419,7 +423,7 @@ export class WebVitalsInstrumentation extends InstrumentationAbstract {
 
   onReportCLS = (
     cls: CLSMetricWithAttribution,
-    applyCustomAttributes?: ApplyCustomAttributesFn,
+    applyCustomAttributes: ApplyCustomAttributesFn = noop,
```
pkanal, line 426: Do we want the default to be a noop here? If that's the case, we'll always call this function on every report and that feels like a step further than checking an if statement. I would rather have it as an optional and do the branch check to avoid the function call everytime but if there is some context I'm missing let me know!

wolfgangcodes: Backed out here 51a94f7, we can revisit in the future :awesome-cat:

pkanal (approved):

PR #282, opened 2024-09-06
chore: Use WeakMap instead of modifying the event.

wolfgangcodes:
Use WeakMap instead of modifying the event for expirmental user-instrumentation.

Use WeakMap instead of modifying the event for expirmental user-instrumentation.

THe example app should emit spans.

### DIFF
--- a/packages/experimental-opentelemetry-web/src/user-interaction-instrumentation.ts
+++ b/packages/experimental-opentelemetry-web/src/user-interaction-instrumentation.ts
@@ -27,7 +27,7 @@ export class UserInteractionInstrumentation extends InstrumentationBase {
   protected _config: UserInteractionInstrumentationConfig;
   private _isEnabled: boolean;
   private _listeners: Listener[];
-  private static _eventMap: WeakMap<Event, Span>;
+  private static _eventMap: WeakMap<Event, Span> = new WeakMap();
 
   constructor(config: UserInteractionInstrumentationConfig = {}) {
     super(INSTRUMENTATION_NAME, VERSION, config);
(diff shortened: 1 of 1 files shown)


packages/experimental-opentelemetry-web/src/user-interaction-instrumentation.ts
```diff
@@ -37,10 +37,53 @@ export class UserInteractionInstrumentation extends InstrumentationBase {
     // enable() gets called by our superclass constructor
     // @ts-expect-error this may get set in enable()
     this._listeners = this._listeners ?? [];
+    UserInteractionInstrumentation._eventMap = new WeakMap();
```
beekhc, line 40: Setting a static member from inside the constructor means that the map will be reset every time you instantiate a new `UserInteractionInstrumentation` instance, right? Is that a problem?

I assume you don't want to assume `UserInteractionInstrumentation` is a singleton, or else you wouldn't need to use a `static` in the first place.

wolfgangcodes: Pragmatically no since instrumentation should only get initialized on startup, BUT def not intended. Thanks for catching! I def forgot to clean this bit up from an earlier code-thought.

wolfgangcodes: See 1ced6b4

packages/experimental-opentelemetry-web/src/user-interaction-instrumentation.ts
```diff
@@ -41,6 +41,47 @@ export class UserInteractionInstrumentation extends InstrumentationBase {
 
   protected init() {}
 
+  private static handleEndSpan(this: void, ev: Event): void {
+    UserInteractionInstrumentation._eventMap.get(ev)?.end();
+  }
+
+  private static createGlobalEventListener(
```
arriIsHere, line 48: ⭐ Making this a static method make a lot more sense.

arriIsHere (commented):
Smoke tested locally and seems to be working well!

arriIsHere (approved):

beekhc (approved):
I don't know this codebase very well, but looks good to me.

PR #303, opened 2024-09-19
feat: support multiple exporters

arriIsHere:
We have had multiple requests to add custom exporters (plural) rather than just the one. This adds a field to the honeycomb SDK configuration that allows providing an array of exporters.

- Closes #<enter issue here>

This adds the `traceExporters` field which takes an array of exporters, it works in the same way that the `traceExporter` field except this takes an array of exporters rather than just the one. 

Smoke test is a WIP but we do have an example project `hello-world-custom-exporter` that shows how to specify multiple exporters including the default one.


packages/honeycomb-opentelemetry-web/examples/hello-world-custom-exporter/index.js
```diff
@@ -0,0 +1,77 @@
...
+
+const configDefaults = {
+  ignoreNetworkEvents: true,
+};
+
+const main = () => {
+  const hcoConfig = {
+    // defaults to sending to US instance of Honeycomb
+    // endpoint: "https://api.eu1.honeycomb.io/v1/traces", // uncomment to send to EU instance
+    apiKey: 'api-key-goes-here',
```
arriIsHere, line 18: only the endpoint, apiKey and dataset (classic environments) fields are used to create the default exporter.

packages/honeycomb-opentelemetry-web/examples/hello-world-custom-exporter/index.js
```diff
@@ -0,0 +1,77 @@
...
+      }),
+    ],
+    contextManager: new ZoneContextManager(),
+    webVitalsInstrumentationConfig: {
+      vitalsToTrack: ['CLS', 'FCP', 'FID', 'INP', 'LCP', 'TTFB'],
+    },
+    traceExporters: [
+      // comment out to only use the console exporter
+      getDefaultTraceExporter(hcoConfig),
+      new ConsoleSpanExporter(),
```
arriIsHere, line 41: One thing I toyed with a bit was just always including the default exporter. It seems some people want the ability to override it so I added this as a way to. But I could also see another config option, such as `excludeDefaultExporter` that defaults to false.

pkanal:
Overall the code looks great! The only thing I want to bring up is I don't love the implicit disabling of the default exporter. I would rather the case where we have another option to explicitly disable the default exporter. OR we override the default exporter with the `traceExporter` option. I think a common use case will be adding exporters on top of the default exporter through the `traceExporters` option, so it is important that that option does not override the default export. (e.g. adding a console exporter or another OTLP exporter to send to a second source).

arriIsHere:
> OR we override the default exporter with the `traceExporter` option. I think a common use case will be adding exporters on top of the default exporter through the `traceExporters` option, so it is important that that option does not override the default export. (e.g. adding a console exporter or another OTLP exporter to send to a second source).

I kinda like this idea personally. It seems more elegant. Only problem I can see is if someone is already using the `traceExporter` field expecting it to be in addition to the default one it would be a breaking change.

packages/honeycomb-opentelemetry-web/examples/hello-world-custom-exporter/index.js
arriIsHere: We discussed this and we are simply moving forward with the list of trace exporters and will add overriding the default exporter in a new PR.

arriIsHere:
We are just going to add multiple exporters in this PR. A followup one will address the override of the default one.

pkanal (approved):
🎉

packages/honeycomb-opentelemetry-web/examples/hello-world-custom-exporter/index.js
wolfgangcodes: It might be a clearer example to move this inline with other config option?

packages/honeycomb-opentelemetry-web/src/types.ts
```diff
@@ -88,6 +88,12 @@ export interface HoneycombOptions extends Partial<WebSDKConfiguration> {
    */
   spanProcessors?: SpanProcessor[];
 
+  /** Provide an array of exporters
+   * Use this to configure custom tracing services to export data in their own format
+   * E.g. You want to send data to another service.
+   */
+  traceExporters?: SpanExporter[];
+
```
wolfgangcodes, line 96: It might be good to call out in the comment that this is in _addition_ to the default exporter. One thing we talked about offline, was a config name that more clearly telegraphs the intent. Do folks have strong objections to `additionalTraceExporters`  to highlight that this is for providing more functionality?

packages/honeycomb-opentelemetry-web/examples/hello-world-custom-exporter/index.js
```diff
@@ -0,0 +1,70 @@
...
+        '@opentelemetry/instrumentation-xml-http-request': configDefaults,
+        '@opentelemetry/instrumentation-fetch': configDefaults,
+        '@opentelemetry/instrumentation-document-load': configDefaults,
+      }),
+    ],
+    contextManager: new ZoneContextManager(),
+    webVitalsInstrumentationConfig: {
+      vitalsToTrack: ['CLS', 'FCP', 'FID', 'INP', 'LCP', 'TTFB'],
+    },
+    traceExporters: [new ConsoleSpanExporter()],
```
wolfgangcodes, line 35: We might want to update the `SDK Configuration` of the readme?

wolfgangcodes (approved):
A few smol comments but nothing show stopping! 🥳  
I would love for us to consider a name for the config option that telegraph the intent more clearly and maybe update the readme since we're adding a new option.

packages/honeycomb-opentelemetry-web/src/types.ts
arriIsHere: Oooo good catch.

PR #309, opened 2024-09-26
feat(honeycomb-opentelemetry-web): Add data attributtes for LCP.

### LINKED ISSUE #95: [feat] Web Vitals Instrumentation data attribute ID
As outlined in @ahrbnsn's [docs PR](https://github.com/honeycombio/honeycomb-opentelemetry-web/pull/61/commits/c48961a76c764c035612a0fcda3b5ef3a2ec8106#diff-4de1aeef841732cd212ead106530bc7bb6f0aa4ed9ac950f4705f0b84fc30d9cR304), it would be nice to grab a human readable data attribute for the element attribution whenever it is available.

- Provide a attribute name as config
- See if the target element has that `data-` attribute
- If it does, add it as an attribute on the span

**Example**
INP target is:
```html
<input type="color" class="css-1rjml27" data-human-id="favorite-color-picker" />
``` 

Config:
``` js
{
  elementDataAttribute: "human-id"
}
``` 

Span attributes:
``` js
{
  ...other attrs
  "inp.element_data_human_id": "favorite-color-picker"
}
```

wolfgangcodes:
Adds `data-` attributes to LCP web vitals and a configuration to specify which data attributes to collect. 

Closes #95 

Add a `data-` to the LCP element and see it's value as an attribute span under `element.data-ATTRIBUTE_NAME`

CleanShot 2024-09-26 at 16 38 09@2x


packages/honeycomb-opentelemetry-web/src/web-vitals-autoinstrumentation.ts
```diff
@@ -88,11 +88,18 @@ interface VitalOpts extends ReportOpts {
   applyCustomAttributes?: ApplyCustomAttributesFn;
 }
 
-interface VitalOptsWithTimings extends VitalOpts {
+interface LcpVitalOpts extends VitalOpts {
+  /**
+   * Will send the values of these data attributes if they appear on an LCP event
+   */
+  dataAttributes?: string[];
```
wolfgangcodes, line 95: This pattern could be useful if we wanted to add in the ability to collect `data-` from nodes where attributes aren't available and we need to do a dom lookup. Maybe this is a way to let users opt-into something that might impact "performance"?

wolfgangcodes: Ooo. default to sending all and filter down? I like that better for this instance.

packages/honeycomb-opentelemetry-web/src/web-vitals-autoinstrumentation.ts
```diff
@@ -481,6 +482,25 @@ export class WebVitalsInstrumentation extends InstrumentationAbstract {
       [`${attrPrefix}.resource_load_time`]: resourceLoadDuration,
     });
 
+    lcpEntry?.element?.getAttributeNames().forEach((attrName) => {
+      // Skip non data-* attributes
+      if (!attrName.startsWith('data-')) {
```
pkanal, line 487: I think it would be safer / more correct to use the [`dataset` API](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes#javascript_access) in JavaScript rather than iterating over all the attributes if it's available!

wolfgangcodes: OH. Heck yeah!

README.md
```diff
@@ -98,6 +98,7 @@ Pass these options to the HoneycombWebSDK:
 | enabled | optional | boolean | `true` | Where or not to enable this auto instrumentation. |
 | lcp| optional| VitalOpts | `undefined` | Pass-through config options for web-vitals. See [ReportOpts](https://github.com/GoogleChrome/web-vitals?tab=readme-ov-file#reportopts).
 | lcp.applyCustomAttributes| optional| function | `undefined` | A function for adding custom attributes to core web vitals spans.
+| lcp.dataAttributes| optional| `string[]` | `undefined` | an array of data-* attribute names to filter. By default it will send all `data-*` attribute-value pairs with the key `lcp.element.data-someAttr`, `[]` will send none.
```
pkanal, line 101: I don't think this is how the values end up being added as attrs, according to the code each one is an attribute by itself with `.data.` prefix, can we change that here?

pkanal (approved):
Amazing! So exciting to have this in! Left a small comment about the documentation but can ship this after that is addressed!

README.md
wolfgangcodes: 💯 I wanted to make sure we were 👍 with [c3f6422](https://github.com/honeycombio/honeycomb-opentelemetry-web/pull/309/commits/c3f64222a27ce9bc4440f455a690695650371307) before updating all the docs. 😅

PR #324, opened 2024-10-08
feat(codeowners): team rename

martin308:

We are renaming the team! This is needed once we have done the switcher-oo


beekhc (approved):
I guess after we change this, we won't be able to submit any code until we rename the team in GitHub?

wolfgangcodes (approved):

martin308:
@beekhc plan is to hit the rename button in GitHub and then merge these team rename PRs

PR #328, opened 2024-10-15
feat: Allow Disabling the default Exporter

arriIsHere:

We have had a few requests come in asking for the ability to disable the default exporter. This makes it possible through a new property in the configuration.

This adds a new SDK configuration property `disableDefaultTraceExporter` that, when set to true will disable the default exporter.

Provide this boolean as `true` in an example and note no events are sent to honeycomb.

### DIFF
--- a/packages/honeycomb-opentelemetry-web/src/span-processor-builder.ts
+++ b/packages/honeycomb-opentelemetry-web/src/span-processor-builder.ts
@@ -37,14 +37,18 @@ export const configureSpanProcessors = (options?: HoneycombOptions) => {
     honeycombTraceExporters.push(...options.traceExporters);
   }
 
+  // Disable this if a configuration option is present
+  if (options?.disableDefaultTraceExporter !== true) {
+    honeycombTraceExporters.push(
+      configureHoneycombHttpJsonTraceExporter(options),
+    );
+  }
+
   // We have to configure the exporter here because the way the base SDK is setup
   // does not allow having both a `spanProcessor` and `traceExporter` configured.
   honeycombSpanProcessor.addProcessor(
     new BatchSpanProcessor(
-      configureCompositeExporter([
-        configureHoneycombHttpJsonTraceExporter(options),
-        ...honeycombTraceExporters,
-      ]),
+      configureCompositeExporter([...honeycombTraceExporters]),
     ),
   );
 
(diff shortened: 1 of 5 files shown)


MustafaHaddara (approved):
I am also not married to the name but I think this is one of those things where there is no perfect name and we could spend weeks bikeshedding about it. 

The current name is Fine™️ 

I vote we ship it! :shipit:

wolfgangcodes (approved):
This works as advertised (esp in conjunction with `traceExporters` option). I'm curious if it's more inline with exceptions where folks can _set_ the default exporter like this: 

```js
    apiKey: 'someKey // Replace with your Honeycomb Ingest API Key
    serviceName: 'coolApp', 
    defaultExporter: MyDefaultTraceExporter(),
    traceExporters: [ExporterMcExporteryFace(), TraceExporter2()]
    ...
    }
```

And disable it with: 
```js
    apiKey: 'someKey // Replace with your Honeycomb Ingest API Key
    serviceName: 'coolApp', 
    defaultExporter: null,
    traceExporters: [ExporterMcExporteryFace(), TraceExporter2()]
    ...
    }
```

I wonder if @pkanal or @martin308 have any instincts here. 🤔

pkanal:
@wolfgangcodes I'm strongly in favour of the `boolean` option, a `null` option makes me nervous about `null`/`undefined` checks and this feels really lightweight and explicit to me!

pkanal (approved):

arriIsHere:
Only thing I am going to change is `push` to `unshift` so that it behaves the same way it does in the previous code.

packages/honeycomb-opentelemetry-web/src/span-processor-builder.ts
```diff
@@ -37,14 +37,18 @@ export const configureSpanProcessors = (options?: HoneycombOptions) => {
     honeycombTraceExporters.push(...options.traceExporters);
   }
 
+  // Disable this if a configuration option is present
+  if (options?.disableDefaultTraceExporter !== true) {
+    honeycombTraceExporters.unshift(
+      configureHoneycombHttpJsonTraceExporter(options),
+    );
+  }
+
```
wolfgangcodes, line 46: We may want a check/logger statement here so that if folks _disable_ the default exporter AND the don't provide any other exporters, there's at least something that's like `console.warn("You gots no exporters configured.")`

PR #359, opened 2024-10-31
chore(dependabot): Dependabot maintenance

pkanal:
We had installed some new dev dependencies that were causing our dependabot PRs to multiply instead of being grouped together.

- Updated the grouped dev dependencies with newly added dependencies
- Removed unused dev dependencies
- Added dependabot config for examples
- Changes examples dependabot cadence to monthly
- Added a README section about dependabot

I linted the file so we just hope and pray?


.github/dependabot.yml
```diff
@@ -22,39 +23,29 @@ updates:
...
           - '@type*'
           - 'cypress'
           - 'eslint*'
           - 'jest*'
+          - 'msw'
           - 'prettier'
+          - 'rollup-plugin*'
           - 'ts-*'
+          - 'tslib'
           - 'typescript'
```
martin308, line 37:
```suggestion
        dependency-type: "development"
```

Can we just do this instead or do we need/want to specify the groups with a pattern?

martin308 (approved):

.github/dependabot.yml
pkanal: Oh yes this is way better, I completely missed this option 🤦🏽‍♀️

PR #382, opened 2024-11-07
fix(rollup): Remove x64 specific package

martin308:

Removes installing the arch specific package of rollup

the build passes


wolfgangcodes (approved):
Build passes!

PR #383, opened 2024-11-07
feat: Add additional fields collected by the TraceKit package for stack traces

jairo-mendoza:

Introduces the [TraceKit](https://github.com/csnover/TraceKit) package to the web distro. Adds additional fields with the help of the `TraceKit` package. These additional fields are prefixed with `exception.structured_stacktrace`. Additional fields will be sent by `GlobalErrorsInstrumentation`.

I tested locally using the `hello-world-web` example provided by the web-distro. Made the example to throw a simple error. This was the result I received on Honeycomb:

Screenshot 2024-11-07 at 3 34 27 PM

Big thanks to @pkanal for helping with this change!

### DIFF
--- a/packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts
+++ b/packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts
@@ -7,6 +7,7 @@ import {
   SEMATTRS_EXCEPTION_STACKTRACE,
   SEMATTRS_EXCEPTION_TYPE,
 } from '@opentelemetry/semantic-conventions';
+import { computeStackTrace, StackFrame } from 'tracekit';
 
 export interface GlobalErrorsInstrumentationConfig
   extends InstrumentationConfig {}
@@ -26,6 +27,34 @@ export class GlobalErrorsInstrumentation extends InstrumentationAbstract {
     this._isEnabled = enabled;
   }
 
+  _computeStackTrace = (error: Error | undefined) => {
+    if (!error) {
+      return {};
+    }
+
+    // OTLP does not accept arrays of objects
+    // breaking down the stack into arrays of strings/numbers
+    const structuredStack: StackFrame[] = computeStackTrace(error).stack;
+    const lines: number[] = [];
+    const columns: number[] = [];
+    const functions: string[] = [];
+    const urls: string[] = [];
+
+    for (const stackFrame of structuredStack) {
+      lines.push(stackFrame.line);
+      columns.push(stackFrame.column);
+      functions.push(stackFrame.func);
+      urls.push(stackFrame.url);
+    }
+
+    return {
+      'exception.structured_stacktrace.columns': columns,
+      'exception.structured_stacktrace.lines': lines,
+      'exception.structured_stacktrace.functions': functions,
+      'exception.structured_stacktrace.urls': urls,
+    };
+  };
+
   onError = (event: ErrorEvent | PromiseRejectionEvent) => {
     const error: Error | undefined =
       'reason' in event ? event.reason : event.error;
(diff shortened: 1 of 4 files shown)


packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts
```diff
@@ -26,6 +27,34 @@ export class GlobalErrorsInstrumentation extends InstrumentationAbstract {
...
   }
 
+  _computeStackTrace = (error: Error | undefined) => {
+    if (!error) {
+      return {};
+    }
+
+    // OTLP does not accept arrays of objects
+    // breaking down the stack into arrays of strings/numbers
+    const structuredStack: StackFrame[] = computeStackTrace(error).stack;
```
pkanal, line 37: Is there a file key passed on the structured stack trace?

packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts
```diff
@@ -26,6 +27,34 @@ export class GlobalErrorsInstrumentation extends InstrumentationAbstract {
     this._isEnabled = enabled;
   }
 
+  _computeStackTrace = (error: Error | undefined) => {
```
pkanal, line 30: Can we add unit tests for this transformation function?

packages/honeycomb-opentelemetry-web/test/global-errors-instrumentation.test.ts
```diff
@@ -37,6 +37,10 @@ describe('Global Errors Instrumentation Tests', () => {
         'exception.type': 'Error',
         'exception.message': 'Something happened',
         'exception.stacktrace': expect.any(String),
+        'exception.structured_stacktrace.columns': expect.any(Array),
```
pkanal, line 40: Ah I see we have these unit tests but it would be great to see a unit test that tests the specific data transform logic too if possible!

packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts
jairo-mendoza: Did a quick check and `TraceKit` does not give a key called file. I'm thinking their file key is the `url` key we currently have here.

packages/honeycomb-opentelemetry-web/test/global-errors-instrumentation.test.ts
```diff
@@ -60,4 +64,19 @@ describe('Global Errors Instrumentation Tests', () => {
...
   });
+
+  describe('_computeStackTrace', () => {
+    it('should return an empty object if error is undefined', () => {
+      expect(instr._computeStackTrace(undefined)).toEqual({});
+    });
+
+    it('should return an object with structured stack trace information', () => {
+      expect(instr._computeStackTrace(new Error('This is an error'))).toEqual({
+        'exception.structured_stacktrace.columns': expect.any(Array),
```
pkanal, line 75: I'm happy to have this as a follow up PR but it would be great to assert on an actual transformation of a stack track here instead of checking that this returns an array. We would have to mock the `Error.stack` property and with a sample stack trace and then check that the results are the line numbers, column numbers and urls that we expect.

jairo-mendoza: I noticed a weird quirk with `TraceKit`, mocking `Error.stack` resulted in some lines being missing for certain stacks. `Error.originalStack` works perfectly though.

pkanal: Gotcha, I'm gonna approve this as is and we can tackle these tests separately. Can you leave a TODO comment here for that?

pkanal (approved):

packages/honeycomb-opentelemetry-web/test/global-errors-instrumentation.test.ts
pkanal: We can also do this through smoke tests!

jairo-mendoza: Adding the TODO now

PR #390, opened 2024-11-13
feat: Update tests to use mocked stacktrace

jairo-mendoza:
Mocks a stack trace that is passed to our `_computeStackTrace` function so we can test with actual values.

All tests pass!

### DIFF
--- a/packages/honeycomb-opentelemetry-web/test/global-errors-instrumentation.test.ts
+++ b/packages/honeycomb-opentelemetry-web/test/global-errors-instrumentation.test.ts
@@ -26,22 +26,35 @@ describe('Global Errors Instrumentation Tests', () => {
 
   describe('when enabled', () => {
     it('should create a span when an error escapes', async () => {
+      const err = new Error('Something happened');
+      err.stack =
+        '' +
+        '  Error: Something happened\n' +
+        '    at baz (filename.js:10:15)\n' +
+        '    at bar (filename.js:6:3)\n' +
+        '    at foo (filename.js:2:3)\n' +
+        '    at (filename.js:13:1)';
       setTimeout(() => {
-        throw new Error('Something happened');
+        throw err;
       });
       await timers.setTimeout();
 
       const span = exporter.getFinishedSpans()[0];
       expect(span.name).toBe('exception');
       // TODO: Mock a stack trace and test that it returns the correct keys and values
-      expect(span.attributes).toMatchObject({
+      expect(span.attributes).toEqual({
         'exception.type': 'Error',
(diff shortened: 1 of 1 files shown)


packages/honeycomb-opentelemetry-web/test/global-errors-instrumentation.test.ts
```diff
@@ -71,14 +84,65 @@ describe('Global Errors Instrumentation Tests', () => {
...
     it('should return an object with structured stack trace information', () => {
-      expect(instr._computeStackTrace(new Error('This is an error'))).toEqual({
-        'exception.structured_stacktrace.columns': expect.any(Array),
-        'exception.structured_stacktrace.lines': expect.any(Array),
-        'exception.structured_stacktrace.functions': expect.any(Array),
-        'exception.structured_stacktrace.urls': expect.any(Array),
+      const err = new Error('This is an error');
+      err.stack =
+        '' +
+        '  Error: This is an error\n' +
```
jairo-mendoza, line 91: Got stack trace from TraceKit: https://github.com/csnover/TraceKit/blob/7bde6a7a0396e61501d2ab448e0807556ec0e816/spec/tracekit-spec.js#L10

Adjusted a bit

packages/honeycomb-opentelemetry-web/test/global-errors-instrumentation.test.ts
```diff
@@ -26,22 +26,35 @@ describe('Global Errors Instrumentation Tests', () => {
...
   describe('when enabled', () => {
     it('should create a span when an error escapes', async () => {
+      const err = new Error('Something happened');
+      err.stack =
+        '' +
+        '  Error: Something happened\n' +
+        '    at baz (filename.js:10:15)\n' +
+        '    at bar (filename.js:6:3)\n' +
+        '    at foo (filename.js:2:3)\n' +
+        '    at (filename.js:13:1)';
```
jairo-mendoza, line 36: TraceKit seems to be really strict with the format of the stack frames. If this stack frame were formatted as `at filename.js:13:1` then it would be completely ignored. I checked the [regex in the TraceKit package](https://github.com/csnover/TraceKit/blob/7bde6a7a0396e61501d2ab448e0807556ec0e816/tracekit.js#L771) and seems like parenthesis are required.

jairo-mendoza: Just wanted to point it out in case it was a cause for concern

pkanal (approved):
Amazing! 👏🏽

martin308 (approved):

PR #393, opened 2024-11-19
docs: Update readme's to include new errors attributes

jairo-mendoza:
We have added new attributes to `GlobalErrorsInstrumentation`, updated READMEs to reflect that.

Check READMEs.


README.md
```diff
@@ -182,6 +182,19 @@ Static fields are added to the [Resource](https://opentelemetry.io/docs/concepts
 
 Fields that can change during the lifetime of the page are instead added to each span in a [SpanProcessor](https://opentelemetry.io/docs/specs/otel/trace/sdk/#span-processor).
 
+#### GlobalErrorsInstrumentationConfig
```
jairo-mendoza, line 185: Wasn't 100% sure where the attributes would fit in the doc, but I think this makes the most sense. I'm open to moving it though!

packages/honeycomb-opentelemetry-web/README.md
```diff
@@ -201,6 +201,19 @@ Static fields are added to the [Resource](https://opentelemetry.io/docs/concepts
 
 Fields that can change during the lifetime of the page are instead added to each span in a [SpanProcessor](https://opentelemetry.io/docs/specs/otel/trace/sdk/#span-processor).
 
+#### GlobalErrorsInstrumentationConfig
```
jairo-mendoza, line 204: Seems like we have 2 README's that are the same, adding in the new info here too

pkanal (approved):

PR #406, opened 2024-12-06
feat(honeycomb-opentelemetry-web): Add support for differing subdomains for debug link exporter.

wolfgangcodes:
This will enable this to work with EU deployments. 

Infers the ui and auth URL roots from the supplied `endpoint`.

Pointing an example app to `https://api.eu1.honeycomb.io/v1/traces` as suggested in the sample apps should link to EU traces.

### DIFF
--- a/packages/honeycomb-opentelemetry-web/src/console-trace-link-exporter.ts
+++ b/packages/honeycomb-opentelemetry-web/src/console-trace-link-exporter.ts
@@ -8,6 +8,7 @@ import {
 } from './util';
 import {
   FAILED_AUTH_FOR_LOCAL_VISUALIZATIONS,
+  MISSING_FIELDS_FOR_GENERATING_LINKS,
   MISSING_FIELDS_FOR_LOCAL_VISUALIZATIONS,
 } from './validate-options';
 import { DiagLogLevel } from '@opentelemetry/api';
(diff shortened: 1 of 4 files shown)


packages/honeycomb-opentelemetry-web/src/console-trace-link-exporter.ts
```diff
@@ -23,13 +24,45 @@ export function configureConsoleTraceLinkExporter(
...
     apiKey,
     options?.logLevel,
+    authRoot,
+    uiRoot,
   );
 }
 
+export const getUrlRoots = (endpoint = '') => {
+  const url = new URL(endpoint);
+  const subdomainRegex = /(api)([.|-])?(.*?)(\.?)(honeycomb.io)(.*)/;
```
MustafaHaddara, line 39: we'll want to make this change:
```suggestion
  const subdomainRegex = /(api)([.|-])?(.*?)(\.?)(honeycomb\.io)(.*)/;
```

otherwise `honeycombxio` at the end would match

packages/honeycomb-opentelemetry-web/src/honeycomb-otel-sdk.ts
```diff
@@ -28,6 +28,7 @@ export class HoneycombWebSDK extends WebSDK {
         ),
       );
     }
+    ``;
```
martin308, line 31: is this for something?

packages/honeycomb-opentelemetry-web/src/console-trace-link-exporter.ts
```diff
@@ -23,13 +24,45 @@ export function configureConsoleTraceLinkExporter(
...
     apiKey,
     options?.logLevel,
+    authRoot,
+    uiRoot,
   );
 }
 
+export const getUrlRoots = (endpoint = '') => {
+  const url = new URL(endpoint);
+  const subdomainRegex = /(api)([.|-])?(.*?)(\.?)(honeycomb\.io)(.*)/;
```
martin308, line 39: What are we trying to capture with the `(.*)` group at the end?

martin308 (approved):
A question or two inline, otherwise looks good to me 🚀

arriIsHere (approved):
Ran through some test apps against EU hc instance and was able to get events and the SDK seemed happy.

packages/honeycomb-opentelemetry-web/src/honeycomb-otel-sdk.ts
wolfgangcodes: A typo! :D

packages/honeycomb-opentelemetry-web/src/console-trace-link-exporter.ts
wolfgangcodes: I  don't think it's strictly needed, but it would match `v1/traces` at the end of the default `endpoint` value.

packages/honeycomb-opentelemetry-web/src/honeycomb-otel-sdk.ts
wolfgangcodes: 9147e28bd13fdc8d9f650d44aa64f06ee4678db3

PR #431, opened 2025-01-13
feat: Add `recordException` helper function

pkanal:
Currently, error auto-instrumentation works well for uncaught errors but it is also useful to have a helper function that sends an exception span that can be called whenever needed for handled exceptions.

- Added a `recordException` helper function that takes an error and creates an exception span for that error.
- Update deprecated semantic attribute constants
- Consolidate logic to format stack traces 

- Tests pass

### DIFF
--- a/packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts
+++ b/packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts
@@ -47,7 +47,7 @@ export function getStructuredStackTrace(error: Error | undefined) {
 
 export function recordException(
   error: Error,
-  attributes: Attributes,
+  attributes: Attributes = {},
   tracer: Tracer = trace.getTracer(LIBRARY_NAME),
 ) {
   const message = error.message;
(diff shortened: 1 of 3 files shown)


packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts
```diff
@@ -1,14 +1,75 @@
...
-  SEMATTRS_EXCEPTION_STACKTRACE,
-  SEMATTRS_EXCEPTION_TYPE,
+  Attributes,
+  context,
+  SpanStatusCode,
+  trace,
+  Tracer,
+} from '@opentelemetry/api';
+import {
+  ATTR_EXCEPTION_MESSAGE,
```
jairo-mendoza, line 12: just checking, but the attribute keys would remain the same right?

jairo-mendoza (approved):
everything looks great!

packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts
pkanal: yes they're the same! just a different variable name!

PR #434, opened 2025-01-14
maint: Move experimental user-interaction-instrumentation into honeycomb-opentelemetry-web/experimental so we can remove experimental-opentelemetry-web 

wolfgangcodes:
This PR moves the experimental user instrumentation to `honeycomb-opentelemetry-web` under the `experimental/` folder. It additionally created a second named entry point to expose experimental features. You can do: 

```ts
import { UserInteractionInstrumentation } from '@honeycombio/opentelemetry-web/experimental';
import { HoneycombWebSDK } from '@honeycombio/opentelemetry-web';
// ...
    debug: true,
    apiKey: 'api-key-goes-here',
    serviceName: 'hny-web-distro-example:user-interaction-instrumentation',
    instrumentations: [
      getWebAutoInstrumentations({
        '@opentelemetry/instrumentation-xml-http-request': configDefaults,
        '@opentelemetry/instrumentation-fetch': configDefaults,
        '@opentelemetry/instrumentation-document-load': configDefaults,
        '@opentelemetry/instrumentation-user-interaction': {
          enabled: false,
        },
      }),
      new UserInteractionInstrumentation({
        rootNodeId: 'root',
      }),
    ], // add automatic instrumentation
    spanProcessors: [new ReactRouterSpanProcessor({ router })],
  });
  sdk.start();
  ```
  
  This allows us to create a separate space for experimentaion, while not incurring the overhead cost of more complex multipackage monorepo tooling.
  
Changes: 
- Use `bundler` resolver, docs [here](https://www.typescriptlang.org/tsconfig/#moduleResolution)
- use `exports` object to support multiple entry points, docs [here](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#exports)
- Add eslint resolver that supports `exports`
  -  ESLint's import plugin doesn't support package.json `exports`. There's a stale open [issue](https://github.com/import-js/eslint-plugin-import/issues/1810)  to fix the workaround is to use the ESLint [typescript resolver](https://github.com/import-js/eslint-import-resolver-typescript). Not all examples are TS, adding an empty config resolves this for the js-only examples.
- Small copy edits to examples
- Moves

The example apps should run and send data to a configured honeycomb instance. 
The smoke tests should pass

### CODE REVIEW

File: packages/honeycomb-opentelemetry-web/package.json [hunk 1]
```diff
@@ -2,14 +2,19 @@
   "name": "@honeycombio/opentelemetry-web",
   "version": "0.11.0",
   "description": "Honeycomb OpenTelemetry Wrapper for Browser Applications",
-  "main": "./dist/cjs/index.js",
```
wolfgangcodes, line 5: eslint doesn't support `exports` this _may_ cause breaking changes and we should call this out in the release notes. Details in the PR desc.

packages/honeycomb-opentelemetry-web/examples/experimental/user-interaction-instrumentation/src/reactRouterSpanProcessor.ts
```diff
@@ -0,0 +1,48 @@
+import { SpanProcessor } from '@opentelemetry/sdk-trace-base';
+import { Span } from '@opentelemetry/api';
+
+/**
+ * SpanProcessor that adds attributes to spans based on the state of the React Router
+ * Sets the page.route attribute to the generic dynamic route
+ * Records the span as an error if there are errors in the router state (e.g. 404)
+ */
+export class ReactRouterSpanProcessor implements SpanProcessor {
```
MustafaHaddara, line 9: Should this live in this example app? Or should we ship it from our distro?

packages/honeycomb-opentelemetry-web/rollup.config.mjs
```diff
@@ -1,33 +1,17 @@
-import { createRequire } from 'node:module';
 import { DEFAULT_EXTENSIONS } from '@babel/core';
 import analyze from 'rollup-plugin-analyzer';
-import babel from '@rollup/plugin-babel';
+import { babel } from '@rollup/plugin-babel';
 import commonjs from '@rollup/plugin-commonjs';
 import dts from 'rollup-plugin-dts';
-import nodeResolve from '@rollup/plugin-node-resolve';
+import { nodeResolve } from '@rollup/plugin-node-resolve';
```
MustafaHaddara, line 6: Why did these imports change? It looks like the package versions stayed the same...

packages/honeycomb-opentelemetry-web/rollup.config.mjs
```diff
@@ -1,33 +1,17 @@
...
 import typescript from 'rollup-plugin-typescript2';
 import autoExternal from 'rollup-plugin-auto-external';
 
-const require = createRequire(import.meta.url);
-const pkg = require('./package.json');
-
-// get dependencies and peer dependencies as declared in package.json
-const getExternalDepsFromPackageJSON = () => {
```
MustafaHaddara, line 15: Do we not need this anymore? It was unused?

packages/honeycomb-opentelemetry-web/package.json
MustafaHaddara: What happens if we leave this in? Anyone not using a bundler aware of `exports` can't see the `/experimental` entry point?

MustafaHaddara (approved):
left a few questions but LGTM 👍

packages/honeycomb-opentelemetry-web/package.json
wolfgangcodes: Nothing, probably. We can add that back in. `exports` is well-supported (except by eslint's import plugin, apparently).

packages/honeycomb-opentelemetry-web/examples/experimental/user-interaction-instrumentation/src/reactRouterSpanProcessor.ts
wolfgangcodes: GREAT Q! I think we can leave that as a followup discussion, since I don't think it the exports for the package have changes, unless I missed something?

packages/honeycomb-opentelemetry-web/rollup.config.mjs
wolfgangcodes: When I did this originally I wrote this code and then realized I could use `rollup-plugin-auto-external` and _because_ this wasn't being linted, the dead code just was chillin.'

packages/honeycomb-opentelemetry-web/rollup.config.mjs
wolfgangcodes: `@rollup/plugin-node-resolve` and `@rollup/plugin-babel` both have names export AND exports that same thing as the default export. It was a linting violation to alias the default export the same as named export, so we just use the named export instead.

packages/honeycomb-opentelemetry-web/package.json
wolfgangcodes: [81481d0](https://github.com/honeycombio/honeycomb-opentelemetry-web/pull/434/commits/81481d0c25a67ec8465d9f704dda068ef344ab5b)

packages/honeycomb-opentelemetry-web/examples/experimental/user-interaction-instrumentation/src/reactRouterSpanProcessor.ts
MustafaHaddara: i just meant that this looks like a useful span processor that we could ship for our clients to use

wolfgangcodes: 💯 Let's follow up and make that call!

PR #435, opened 2025-01-15
maint: Remove experimental-opentelemetry-web

wolfgangcodes:
Follow on from #434 to remove the extraneous package


RELEASING.md
```diff
@@ -1,91 +1,110 @@
...
+  >
   >For _EACH_ package that needs to be released:
+
 14. Create a tag for the new version:
+
     ```shell
       git tag -a honeycomb-opentelemetry-web-vX.Y.Z -m honeycomb-opentelemetry-web-vX.Y.Z
     ```
+
     - These tags MUST correspond to the accepted tag prefixes (see [Available Tag Prefixes](#available-tag-prefixes)).
```
MustafaHaddara, line 98: are we keeping the github tags the same?

MustafaHaddara (approved):

RELEASING.md
wolfgangcodes: We could bo back to just using the version number, but I don't have a strong feeling here. So, I'm inclined to leave it as is. Open to contrary opinions!

PR #442, opened 2025-01-15
fix: Add guard for null stack from tracekit.

### LINKED ISSUE #430: "is not iterable" error in GlobalErrorsInstrumentation._computeStackTrace
**Versions**
HoneycombWebSDK import in code: `@honeycombio/opentelemetry-web` version 0.10.0
OpenTelemetry span library: `@honeycombio/instrumentation-global-errors` version 0.10.0

**Steps to reproduce**

1. Unable to figure out due to this issue itself

**Additional context**

We recently added HoneycombWebSDK to our Nuxt/Vue 3 app and while it does report many reasonable-looking exceptions, there is one that stands out:

```
TypeError: ae is not iterable
    at GlobalErrorsInstrumentation._computeStackTrace ([REDACTED].js:322:39022)
    at GlobalErrorsInstrumentation.onError ([REDACTED].js:322:39539)
```

Since this is bundled code, I located the original source and it is here: https://github.com/honeycombio/honeycomb-opentelemetry-web/blob/7645a8a338792499f3f5e8a40f6dbc2fcebc1d3d/packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts#L43

This suggests to me that `computeStackTrace(error).stack` is not filled in and indeed, looking through the code for https://github.com/csnover/TraceKit/blob/master/tracekit.js, all paths return `mode`, `name`, and `message` but only successful computations return `stack`

This library should cope with `stack` being missing from TraceKit, otherwise these unknown non-parsable stack traces prevent error instrumentation from being sent

wolfgangcodes:

- Closes https://github.com/honeycombio/honeycomb-opentelemetry-web/issues/430

Adds a guard for `stack === null` from `tracekit`. Also added unit test.

### DIFF
--- a/packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts
+++ b/packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts
@@ -30,7 +30,7 @@ export function getStructuredStackTrace(error: Error | undefined) {
   const functions: string[] = [];
   const urls: string[] = [];
 
-  if (!structuredStack) {
+  if (!Array.isArray(structuredStack)) {
     return {};
   }
 
--- a/packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts
+++ b/packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts
@@ -30,12 +30,16 @@ export function getStructuredStackTrace(error: Error | undefined) {
   const functions: string[] = [];
   const urls: string[] = [];
 
-  for (const stackFrame of structuredStack) {
+  if (!structuredStack) {
+    return {};
+  }
+
+  structuredStack.forEach((stackFrame) => {
     lines.push(stackFrame.line);
     columns.push(stackFrame.column);
     functions.push(stackFrame.func);
     urls.push(stackFrame.url);
-  }
+  });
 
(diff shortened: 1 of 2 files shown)


jairo-mendoza (approved):

packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts
```diff
@@ -30,12 +30,16 @@ export function getStructuredStackTrace(error: Error | undefined) {
   const functions: string[] = [];
   const urls: string[] = [];
 
-  for (const stackFrame of structuredStack) {
+  if (!structuredStack) {
```
pkanal, line 33: Nit: If we want to be extra safe here we might also want to check that `structuredStack` is iterable.

pkanal (approved):

packages/honeycomb-opentelemetry-web/src/global-errors-autoinstrumentation.ts
wolfgangcodes: 💯 
[28a6ac4](https://github.com/honeycombio/honeycomb-opentelemetry-web/pull/442/commits/28a6ac4ab2b50e76b848af52b0c34f13497c0f84)

PR #447, opened 2025-01-21
feat(error-autoinstrumentation): Add applyCustomAttribitesToSpan config option

pkanal:
Generally auto-instrumentation should include a callback config option that allows mutating the span (to add attributes / whatever else) which was missing in our config. This PR adds the `applyCustomAttribitesToSpan` that exposes the error object and the span through a callback function allowing the user to mutate each error span in the config. This option is also on the `recordException` function. Ideally we will eventually deprecate the `attributes` argument on this function but that would be a breaking change right now so I'll leave that for a future PR.

- Adds `applyCustomAttribitesToSpan` config option to error instrumentation
- Adds `applyCustomAttribitesToSpan` argument to `recordException` function
- Docs
- Tests

- Tests pass


wolfgangcodes (approved):

martin308 (approved):
mm attribites :chompy:

PR #468, opened 2025-02-27
chore: remove unused gh action

MustafaHaddara:
This action started failing

Removes the failing action (we weren't using it anyways)

- [x] the build passes


JamieDanielson (approved):
I don't think my review is sufficient for passing checks but signaling my support for this change 😄

beekhc (approved):

PR #477, opened 2025-03-06
fix: Dont validate options when using a custom collector.

### LINKED ISSUE #476: Spurious warning when supplying `endpoint`
We seem to get an inaccurate warning when specifying `endpoint`. 

> @wolfgangcodes we do get this warning now though:
> 
> Image
> 
> This is in a scenario in which we don't define an `apiKey` option, but define an `endpoint` option instead.
> 
> We see the traces in Honeycomb though, even if we get a warning saying we won't. 

 _Originally posted by @patricklafrance in [#443](https://github.com/honeycombio/honeycomb-opentelemetry-web/issues/443#issuecomment-2683304255)_

beekhc:

- Closes #476

This changes options validation so that we skip requiring all options when using a custom collector. We consider any endpoint that doesn't end with `".honeycomb.io"` to be a custom collector.

New unit tests have been added.
Smoke tests still pass.


martin308 (approved):
Nice!

beekhc:
Thanks for the quick review!

But I want to run it by the team tomorrow before I merge, in case it's controversial.

wolfgangcodes (approved):

arriIsHere (approved):

PR #486, opened 2025-03-20
maint: Consolidate all the dependabot for examples

martin308:
While investigating why we aren't getting dependabot PRs made for the example apps I consolidated all these together to make it easier to reason about.

I can't think of a reason that we should have them separate?


pkanal (approved):
oh yes amazing, thank you!

PR #499, opened 2025-03-28
Add top-level types element to package.json

sodabrew:

Errors reporting unable to find types for this package. Here is the error I get when running under `jest`

``` 
src/javascript/tracing/honeycomb.ts:4:8 - error TS7016: Could not find a declaration file for module '@honeycombio/opentelemetry-web'. 'node_modules/@honeycombio/opentelemetry-web/dist/cjs/index.js' implicitly has an 'any' type.
      Try `npm i --save-dev @types/honeycombio__opentelemetry-web` if it exists or add a new declaration (.d.ts) file containing `declare module '@honeycombio/opentelemetry-web';`

    4 } from '@honeycombio/opentelemetry-web'
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This PR resolves the error.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1209957762639823

### DIFF
--- a/packages/honeycomb-opentelemetry-web/package.json
+++ b/packages/honeycomb-opentelemetry-web/package.json
@@ -3,6 +3,7 @@
   "version": "0.15.0",
   "description": "Honeycomb OpenTelemetry Wrapper for Browser Applications",
   "main": "./dist/cjs/index.js",
+  "types": "./dist/types/index.d.ts",
   "exports": {
     ".": {
       "types": "./dist/types/index.d.ts",
@@ -13,6 +14,7 @@
     },
     "./experimental": {
       "types": "./dist/types/experimental/index.d.ts",
+      "node": "./dist/cjs/experimental/index.js",
       "require": "./dist/cjs/experimental/index.js",
       "import": "./dist/esm/experimental/index.js",
       "default": "./dist/esm/experimental/index.js"


sodabrew:
@pkanal hoping you'll take a look at this PR soon, thanks!

pkanal:
> @pkanal hoping you'll take a look at this PR soon, thanks!

@sodabrew appreciate your patience! Will take a look this week.

sodabrew:
Turns out I was using a too-old version of Jest. See https://github.com/open-telemetry/opentelemetry-js/issues/5618

tl;dr: use Jest v29.4 or higher because jest-resolve module loader supports the types from imports/exports (don't need this PR) and submodule imports (resolves the issue I was seeing in the 5618 issue linked above)

Murphy's Law dictated that once I wrote the issue down as a ticket I would find a solution a few minutes later (followed by a couple hours of actually implementing the solution, but, you know, _engineering_ the solution.)

sodabrew:
Reopening. Yes, a newer version of Jest will solve this problem. But as I found and documented a workaround so that I don't need to upgrade Jest yet -- it's a huge lift -- then I still need this change to be in this package.

sodabrew:
Aaaand here's the next workaround! Not blocked:

```
jest.config.js:

+  globals: {
+    'ts-jest': {
+      useESM: true
+    }
+  },

```

packages/honeycomb-opentelemetry-web/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.15.0",
   "description": "Honeycomb OpenTelemetry Wrapper for Browser Applications",
   "main": "./dist/cjs/index.js",
+  "types": "./dist/types/index.d.ts",
```
sodabrew, line 6: This fixes the type import problem I am experiencing while using Jest v27

pkanal (approved):
Thanks for adding this and following up! This change looks good to me 👍🏽

sodabrew:
Thank you!

PR #510, opened 2025-04-07
feat: add an option to disable automatic browser attributes

beekhc:

It is currently not possible to disable automatic attributes inferred by the browser, which is inconsistent with the options to turn off other auto-instrumentation. It also prevents this SDK from being used in non-browser web contexts.

Adds a new `disableBrowserAttributes` config option, which defaults to false. Turning on this option prevents the Honeycomb SDK from automatically adding browser-derived resource attributes, as well as the browser attribute span processor.

- New unit tests added.
- Smoke tests still pass.

### DIFF
--- a/packages/honeycomb-opentelemetry-web/src/configure-resource-attributes.ts
+++ b/packages/honeycomb-opentelemetry-web/src/configure-resource-attributes.ts
@@ -15,9 +15,13 @@ export const configureResourceAttributes = (
   options?: HoneycombOptions,
 ): Resource => {
   let resource = resourceFromAttributes({})
-    .merge(configureEntryPageResource(options?.entryPageAttributes))
-    .merge(configureBrowserAttributesResource())
-    .merge(configureHoneycombResource());
+    .merge(configureEntryPageResource(options?.entryPageAttributes));
+
+  if (!options?.disableBrowserAttributes) {
+    resource = resource.merge(configureBrowserAttributesResource())
+  }
+
+  resource = resource.merge(configureHoneycombResource());
 
   if (options?.resource) {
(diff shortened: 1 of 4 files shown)


packages/honeycomb-opentelemetry-web/test/honeycomb-otel-sdk.test.ts
```diff
@@ -226,3 +227,54 @@ describe('span processor config', () => {
     });
   });
 });
+
+describe('disabling browser attributes', () => {
```
pkanal, line 231: Love the tests!

packages/honeycomb-opentelemetry-web/src/configure-resource-attributes.ts
```diff
@@ -14,10 +14,15 @@ import { configureBrowserAttributesResource } from './browser-attributes-resourc
 export const configureResourceAttributes = (
   options?: HoneycombOptions,
 ): Resource => {
-  let resource = resourceFromAttributes({})
-    .merge(configureEntryPageResource(options?.entryPageAttributes))
-    .merge(configureBrowserAttributesResource())
-    .merge(configureHoneycombResource());
+  let resource = resourceFromAttributes({}).merge(
+    configureEntryPageResource(options?.entryPageAttributes),
```
pkanal, line 18: I think the entry page attributes should also be included in browser attributes

beekhc: Done. And test updated.

I had missed this one because it's designed to skip adding them if `window` is undefined.

pkanal (approved):

PR #517, opened 2025-04-23
fix: explicitly add default resource attributes

pkanal:
As of v0.15.0 (the version where we upgraded to v2 of OTel packages), the following attributes disappeared from telemetry:
- `telemetry.sdk.name`
- `telemetry.sdk.language`
- `telemetry.sdk.version`

It seems like in v2, these have to be [explicitly merged with other resources](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-sdk-trace-base/src/BasicTracerProvider.ts#L49) when using the `WebTracerProvider` because the default behaviour is to overwrite them with resources that have already been passed in.

- Add `defaultResource` to base SDK
- Add unit tests to make sure these attributes are present
- Add smoke tests to make sure these attributes are present

- Check example apps resource attributes
- Unit tests pass
- Smoke tests pass


beekhc (approved):
🚀

PR #545, opened 2025-06-02
feat: add metrics and logs configuration

beekhc:

Adds configuration for metrics and logs to Honeycomb Web SDK.

* This adds configuration of metrics and logs analogous to what we've got in the mobile SDKs.
* I also had to add a `forceFlush` method to the SDK to get it to send the metrics for the smoke tests.

See the new smoke tests.


packages/honeycomb-opentelemetry-web/src/util.ts
```diff
@@ -71,16 +75,52 @@ export const getTracesEndpoint = (options?: HoneycombOptions) => {
...
   // use `endpoint` option if provided and append '/v1/traces' if not already appended
   if (options?.endpoint) {
-    return maybeAppendTracesPath(options.endpoint);
+    return maybeAppendPath(options.endpoint, TRACES_PATH);
   }
 
   return DEFAULT_TRACES_ENDPOINT;
 };
 
+export const getMetricsEndpoint = (options?: HoneycombOptions) => {
```
pkanal, line 84: Can we add unit tests for this similar to `getTracesEndpoint`? Even though its a bit tedious it really does help to keep the coverage high for when we make changes to these functions in the future.

packages/honeycomb-opentelemetry-web/src/util.ts
```diff
@@ -71,16 +75,52 @@ export const getTracesEndpoint = (options?: HoneycombOptions) => {
...
+
+  // use `endpoint` option if provided and append '/v1/metrics' if not already appended
+  if (options?.endpoint) {
+    return maybeAppendPath(options.endpoint, METRICS_PATH);
+  }
+
+  return DEFAULT_METRICS_ENDPOINT;
+};
+
+export const getLogsEndpoint = (options?: HoneycombOptions) => {
```
pkanal, line 98: Same comment here about unit tests!

pkanal:
> The unit tests are failing because the browser attribute tests require a browser environment, but otel is importing the node platform version of `unrefTimer`. Do you have any ideas how to fix this?

Hmmmm I think this is because the tests technically run in a node environment, the first thing I would try is making sure that the [test env is JSDOM](https://jestjs.io/docs/configuration#testenvironment-string) I think we can do that for a single file as outlined in the example there. If that doesn't work we might need to mock it or figure something else out.

pkanal:
> > > The unit tests are failing because the browser attribute tests require a browser environment, but otel is importing the node platform version of `unrefTimer`. Do you have any ideas how to fix this?
> > 
> > 
> > Hmmmm I think this is because the tests technically run in a node environment, the first thing I would try is making sure that the [test env is JSDOM](https://jestjs.io/docs/configuration#testenvironment-string) I think we can do that for a single file as outlined in the example there. If that doesn't work we might need to mock it or figure something else out.
> 
> I tried that, but the test was already `jsdom`. I tried changing it to `node` and even _more_ tests failed, because `document` is undefined, which breaks the browser attributes.

Then I think our best bet is to [mock `unrefTimer` in the setup file for jest](https://github.com/honeycombio/honeycomb-opentelemetry-web/blob/main/packages/honeycomb-opentelemetry-web/setup-jest.ts). I've had to do it in some projects with OTel + tests and it isn't ideal but it is the best we can do for now.

beekhc:
> > > > The unit tests are failing because the browser attribute tests require a browser environment, but otel is importing the node platform version of `unrefTimer`. Do you have any ideas how to fix this?
> > > 
> > > 
> > > Hmmmm I think this is because the tests technically run in a node environment, the first thing I would try is making sure that the [test env is JSDOM](https://jestjs.io/docs/configuration#testenvironment-string) I think we can do that for a single file as outlined in the example there. If that doesn't work we might need to mock it or figure something else out.
> > 
> > 
> > I tried that, but the test was already `jsdom`. I tried changing it to `node` and even _more_ tests failed, because `document` is undefined, which breaks the browser attributes.
> 
> Then I think our best bet is to [mock `unrefTimer` in the setup file for jest](https://github.com/honeycombio/honeycomb-opentelemetry-web/blob/main/packages/honeycomb-opentelemetry-web/setup-jest.ts). I've had to do it in some projects with OTel + tests and it isn't ideal but it is the best we can do for now.

I will try that.

Seeing how [OTel already mocks it in the `browser` context](https://github.com/open-telemetry/opentelemetry-js/blob/56df3c03468303d7f7767ff5244c9e190144082b/packages/opentelemetry-core/package.json#L8), I tried changing `jest.config.js` from:
```
  testEnvironmentOptions: {
    customExportConditions: [''],
  },
```
to
```
  testEnvironmentOptions: {
    customExportConditions: ['browser'],
  },
```

I was hoping that might make it import the right version. That didn't fix it, and also broke another test that depends on some node imports. ಠ_ಠ

beekhc:
> > > > > The unit tests are failing because the browser attribute tests require a browser environment, but otel is importing the node platform version of `unrefTimer`. Do you have any ideas how to fix this?
> > > > 
> > > > 
> > > > Hmmmm I think this is because the tests technically run in a node environment, the first thing I would try is making sure that the [test env is JSDOM](https://jestjs.io/docs/configuration#testenvironment-string) I think we can do that for a single file as outlined in the example there. If that doesn't work we might need to mock it or figure something else out.
> > > 
> > > 
> > > I tried that, but the test was already `jsdom`. I tried changing it to `node` and even _more_ tests failed, because `document` is undefined, which breaks the browser attributes.
> > 
> > 
> > Then I think our best bet is to [mock `unrefTimer` in the setup file for jest](https://github.com/honeycombio/honeycomb-opentelemetry-web/blob/main/packages/honeycomb-opentelemetry-web/setup-jest.ts). I've had to do it in some projects with OTel + tests and it isn't ideal but it is the best we can do for now.
> 
> I will try that.
> 
> Seeing how [OTel already mocks it in the `browser` context](https://github.com/open-telemetry/opentelemetry-js/blob/56df3c03468303d7f7767ff5244c9e190144082b/packages/opentelemetry-core/package.json#L8), I tried changing `jest.config.js` from:
> 
> ```
>   testEnvironmentOptions: {
>     customExportConditions: [''],
>   },
> ```
> 
> to
> 
> ```
>   testEnvironmentOptions: {
>     customExportConditions: ['browser'],
>   },
> ```
> 
> I was hoping that might make it import the right version. That didn't fix it, and also broke another test that depends on some node imports. ಠ_ಠ

I tried mocking `setInterval` / `setTimeout` to return something that would work, but I couldn't get it to work and make the type system happy. So, instead, I changed jest's resolver to support the `browser` attribute in `package.json`, the same as we had to do to make the react native sample app work. I think this is the least bad solution, but let me know if you think it's a problem.

Also, I added all the unit tests you requested.

pkanal (approved):
🎉

PR #534, opened 2025-05-13
chore: Upgrade to web-vitals 5.0.0

wolfgangcodes:
## Which problem is this PR solving?

- Closes https://github.com/honeycombio/honeycomb-opentelemetry-web/issues/531

## Short description of the changes
Upgrades web-vitals to `5.0.0` this removes `FID` from the CWV since it was replaced by INP.

The full change log details are here:
https://github.com/GoogleChrome/web-vitals/blob/main/docs/upgrading-to-v5.md
and 
https://github.com/GoogleChrome/web-vitals/blob/main/CHANGELOG.md#v500-2025-05-07

PR #548, opened 2025-06-05
chore: Update examples to remove FID.

wolfgangcodes:
FID which was removed in #534 was still referenced in examples. This updates the readme and examples.


beekhc (approved):
9wwpky

PR #554, opened 2025-06-13
feat: Send data-* attributes for INP.

wolfgangcodes:

Adds support to send `data-*` attributes for INP.

Additionally updates the `hello-world-web` to have additional `data-`` and furnished an example for a custom session Id provider.

Add for INP entries to span `inp` spans.

Unit tests and observing sample app the span with debug set to true, should show the data-* attributes attached


packages/honeycomb-opentelemetry-web/src/web-vitals-autoinstrumentation.ts
```diff
@@ -66,6 +66,11 @@ interface InpVitalOpts extends VitalOpts {
    * if this is true it will create spans from the PerformanceLongAnimationFrameTiming frames
    */
   includeTimingsAsSpans?: boolean;
+  /**
+   * Will filter the values of these data attributes if provided, otherwise will send all data-* attributes an INP entry
+   * An empty allow list, such as { dataAttributes: [] } will disable sending data-* attributes
+   */
+  dataAttributes?: string[];
```
wolfgangcodes, line 73: Update the readme.

martin308 (approved):
looks good. I'm not sure how the session id example fits in with this feature? Should it be another PR?

wolfgangcodes:
> looks good. I'm not sure how the session id example fits in with this feature? Should it be another PR?

It could, if we feel strongly about pulling it into it's own PR, I updated the desc. 
It was a useful addition for testing when manually verifying.

PR #571, opened 2025-07-16
feat: Export attribute names as constants

pkanal:
Exports all attribute names introduced by the distro as constants to be used both internally and externally by users.

- added a semantic-attributes.ts file with every attribute name as a constant similar to the OpenTelemetry semantic conventions attribute files
- used these constants internally in the distro instead of hard coded strings.

- tests pass
- smoke tests pass

### DIFF
--- a/packages/honeycomb-opentelemetry-web/src/semantic-attributes.ts
+++ b/packages/honeycomb-opentelemetry-web/src/semantic-attributes.ts
@@ -1,0 +1,673 @@
+/**
+ * Custom semantic attribute constants for the Honeycomb OpenTelemetry Web SDK.
+ * These attributes extend the standard OpenTelemetry semantic conventions with
+ * Honeycomb-specific and browser-specific attributes.
+ */
+
+// =============================================================================
+// Browser Attributes
+// =============================================================================
+
+/**
+ * The name of the browser.
+ * @example "Chrome", "Firefox", "Safari"
+ */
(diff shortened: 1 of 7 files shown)


packages/honeycomb-opentelemetry-web/src/semantic-attributes.ts
wolfgangcodes, line 1: [suggestion] Can we break this file out by concern? So it's easier to see which ones are honeycomb centric and which ones are filling gaps upstream? They already seem to have a relevant grouping!

packages/honeycomb-opentelemetry-web/src/semantic-attributes.ts
```diff
@@ -0,0 +1,673 @@
...
+
+// =============================================================================
+// Browser Attributes
+// =============================================================================
+
+/**
+ * The name of the browser.
+ * @example "Chrome", "Firefox", "Safari"
+ */
+export const ATTR_BROWSER_NAME = 'browser.name';
```
wolfgangcodes, line 15: [THought] Would it be useful to maybe link out to possible registry definitions in a readme or the headers with a blurb about how we care filling gaps 
https://github.com/open-telemetry/semantic-conventions/tree/main/model/browser

wolfgangcodes (approved):
A few smol non-blocking suggestions!

PR #601, opened 2025-09-11
docs: Add documentation for sending data to EU instance

arriIsHere:
By default our SDK will send data to the North America instance. As many customers use the European Union instance we need documentation on how to send data there.

- Closes #<enter issue here>

Adds an aside in `getting started` with information on setting the `endpoint` property to the eu instance.


wolfgangcodes (approved):

wolfgangcodes (commented):
I think we were intending to add color to our configurations snippet as well:

```
import { HoneycombWebSDK, WebVitalsInstrumentation } from '@honeycombio/opentelemetry-web';
import { getWebAutoInstrumentations } from '@opentelemetry/auto-instrumentations-web';

  apiKey: 'api-key-goes-here',
  serviceName: 'your-great-browser-application',
  instrumentations: [getWebAutoInstrumentations(), new WebVitalsInstrumentation()], // add automatic instrumentation
});
sdk.start();
```

So that it has the same info as: https://docs.honeycomb.io/get-started/start-building/application/web/

PR #602, opened 2025-09-14
fix: fix instrumentation to get correct tracer

beekhc:

Before this change, Instrumentation is initialized before the default tracer is constructed, so that the instrumentation's `tracer` property won't work. This fixes it.

1. Set the default tracer as the global tracer.
2. Move instrumentation initialization until after that tracer is created.

* Smoke tests still pass.
* I manually tested with the React Native repo that the `tracer` property on instrumentation fails before and works after this change.


wolfgangcodes (approved):
🚀

arriIsHere (approved):

PR #622, opened 2025-11-19
fix: Remove console log from onError handler

msolli:

- Closes #621

Removes stray console log statement from onError function.

Throw an error in an instrument application and observe that no `undefined` value is logged.


wolfgangcodes (approved):
Thank you!

PR #656, opened 2026-08-13
feat: report Core Web Vitals for soft navigations

wolfgangcodes:

Core Web Vitals are only measured for the first page load. Single-page apps change route without reloading the document, so those route changes are never measured. `web-vitals` v6 adds support for them ("soft navigations").

The upgrade was blocked. The `before=2025-09-08` pin in `.npmrc` (added in #608 after shai-hulud) blocks every `web-vitals` release after 5.1.0. That pin was a stopgap until npm shipped npm/cli#8570, which now exists as `min-release-age`.

- Upgrade `web-vitals` 5.1.0 → 6.1.0. The `on*()` signatures are unchanged, so no call sites changed.
- Report vitals for soft navigations, opt in per vital.
- Anchor spans to `navigationStartTime` rather than the page's time origin. `web-vitals` reports `0` there for hard navigations, so **hard-navigation timings do not change**. FCP takes its end time from `navigationStartTime + value`, since `web-vitals` attributes no `fcpEntry` for a soft navigation. TTFB collapses onto the navigation start, since a soft navigation sends no request.
- Add `navigation_id`, `navigation_url`, `navigation_start_time` and `navigation_interaction_id` to every vital. `navigation_type` can now also be `soft-navigation`. Prefer `navigation_url` over the URL at export time: a metric can be reported after the next navigation begins.
- Add route buttons to the `hello-world-custom-exporter` example.
- Replace the `before` pin with `min-release-age=7` in all 8 `.npmrc` files. `before` overrides `min-release-age` within one file, so it had to be removed rather than supplemented.

`web-vitals` is a bundled `dependency`, so this is **not** a breaking change for SDK users.

## New configuration

Soft navigation reporting is off by default. Opt in per vital. Requires Chromium 151+; other browsers ignore the flag, so setting it everywhere is safe.

```js
webVitalsInstrumentationConfig: {
  cls: { reportSoftNavs: true },
  inp: { reportSoftNavs: true },
  lcp: { reportSoftNavs: true },
}
```

`min-release-age` needs npm >= 11.10.0, so Node is pinned in `.nvmrc` (24.19.0, which bundles npm 11.17.0) and CI builds on `cimg/node:24.19.0`. npm only warns about unknown config keys, so CI fails if npm does not understand it. **Contributors on an older npm lose the quarantine.** Check with:

```sh
cd packages/honeycomb-opentelemetry-web && npm config get min-release-age
```

Watch for an `Unknown project config` warning, not the value — `npm config get` prints `7` either way.

```sh
cd packages/honeycomb-opentelemetry-web
npm ci && npm run test:ci && npm run build && npm run lint:ci && npm run check-format
```

167 tests pass, 1 pre-existing skip. Six new tests cover the soft-nav paths: the `reportSoftNavs` hand-off to `web-vitals`, FCP/CLS/LCP/INP anchoring to the navigation start, TTFB collapsing onto it, and `navigation_url`.

Soft navigations need real Chrome, so CI does not cover those paths. They were checked manually in Chrome 151, where hard-navigation spans still start at 0 and soft-navigation spans start at the browser's soft-nav `startTime`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


andaley (commented):
excited to be able to have this data. thank you for kicking it off so quickly!

packages/honeycomb-opentelemetry-web/src/web-vitals-autoinstrumentation.ts
```diff
@@ -603,8 +657,16 @@ export class WebVitalsInstrumentation extends InstrumentationAbstract {
     }: INPAttribution = attribution;
 
     const inpDuration = inputDelay + processingDuration + presentationDelay;
-    const startTime = hrTime(interactionTime);
-    const endTime = hrTime(interactionTime + inpDuration);
+    // As of web-vitals v6 `interactionTime` is optional: it is absent when the
```
pkanal: nit: we might want to run a `/clarity-review` on these comments, they're a little verbose.

pkanal (approved):

MustafaHaddara (approved):

PR #660, opened 2026-08-21
maint: update dependencies to the 0.220 OpenTelemetry generation

wolfgangcodes:
Dependency alignment only: **no public API change and no telemetry change**. Ships as a minor.

Updates OpenTelemetry package to the 0.220 and modernises the toolchain, so `@opentelemetry/browser-instrumentation@0.6.0` can be installed alongside this SDK. As well as security fixes and patches to deps.

No API or telemetry changes in this version.

## OpenTelemetry dependencies

| Package | From | To |
|---|---|---|
| `api-logs`, `instrumentation`, `web-common`, `opentelemetry-browser-detector`, the three exporters | `0.203.0` | `0.220.0` |
| `core`, `resources`, `sdk-trace-base`, `sdk-trace-web` | `^2.0.0` | `~2.9.0` |
| `auto-instrumentations-web` (dependency **and** peer) | `^0.49.0` | `^0.65.0` |
| `semantic-conventions` | `^1.30.0` | `^1.43.0` |
| `api` | `~1.9.0` | `~1.9.1` |
| `tracekit` | `^0.4.7` | `^0.4.9` |

The `tracekit` addresses #618 

## Pinned at 0.220.0, not 0.221.0

0.220.0 is the last release that emits **legacy HTTP semantic conventions by default** and still honours `semconvStabilityOptIn`.
See: https://github.com/open-telemetry/opentelemetry-js/releases#release-experimental/v0.221.0. From 0.221.0 the instrumentations emit stable names, which will be a breaking change for this SDK. 

## Verification

- `npm test` — **181 passed, 1 skipped**, 19/19 files
- `build`, `lint`, `check-format`, `typecheck` — all clean
- `npm run example:lint:all` — clean
- `npm ls` — single copy of every OpenTelemetry package

PR #661, opened 2026-08-22
feat: add opt-in navigation and resource timing

wolfgangcodes:
Stacked on #660, which is what makes this possible: `@opentelemetry/browser-instrumentation@0.6.0` pins `instrumentation` and `api-logs` at `^0.220.0`, the generation that PR lands on.

## What this adds

Navigation timing and resource timing, from the browser SDK OpenTelemetry is building at [open-telemetry/opentelemetry-browser](https://github.com/open-telemetry/opentelemetry-browser).

```js
  navigationTimingInstrumentationConfig: { enabled: true },
  resourceTimingInstrumentationConfig: { enabled: true },
});
```

**Both are off by default** because resource timing records an event for *every* asset a page loads. Explicit opt-in allows is to add it as as an opt-in, it is likely to default to true in future verions. 

## They emit events, not spans

Both produce log records, arriving as **`browser.navigation_timing`** and **`browser.resource_timing`** events rather than spans. They therefore depend on the logs pipeline from #658, and pick up `session.id` from #659 once that lands — the three pieces compose.

- `browser.navigation_timing` — one event per page load with the `PerformanceNavigationTiming` milestones: DNS, connect, request, response, DOM interactive, DOM complete, load event, transfer sizes.
- `browser.resource_timing` — one event per resource: scripts, stylesheets, images, fonts, XHR and fetch.

## The feedback loop, and why ignoreUrls is seeded

Resource timing records an event for every resource the browser fetches, **including this SDK's own exports**. 

To prevent a feedback loop, the resolved Honeycomb endpoints are always prepended to `ignoreUrls`, in addition to anything the user-supplied:

```
[ tracesEndpoint, metricsEndpoint, logsEndpoint, ...callerSupplied ]
```

## Verification

- `npm run test:ci` — **193 passed, 1 skipped** (12 new)
- `typecheck`, `lint:ci`, `check-format`, `build` — all clean


packages/honeycomb-opentelemetry-web/src/types.ts
```diff
@@ -241,6 +243,26 @@ export interface HoneycombOptions extends Partial<WebSDKConfiguration> {
...
+
+  /**
+   * Config options for resource timing instrumentation. Disabled by default;
+   * set `enabled: true` to turn it on.
+   *
+   * Emits a `browser.resource_timing` log record per resource the browser
+   * loads, so it is the higher volume of the two. Narrow it with
+   * `initiatorTypes` and `ignoreUrls`. This SDK's own export endpoints are
+   * always ignored.
+   */
```
wolfgangcodes, line 261:
```suggestion
  /**
   * Config options for resource timing instrumentation. Disabled by default;
   * set `enabled: true` to turn it on.
   *
   * Emits a `browser.resource_timing` log record per resource the browser
   * loads. See the types `initiatorTypes` and `ignoreUrls` for configuration options. 
   * This SDK's own export endpoints are always ignored.
   */
```

packages/honeycomb-opentelemetry-web/src/types.ts
```diff
@@ -241,6 +243,26 @@ export interface HoneycombOptions extends Partial<WebSDKConfiguration> {
   webVitalsInstrumentationConfig?: WebVitalsInstrumentationConfig;
   globalErrorsInstrumentationConfig?: GlobalErrorsInstrumentationConfig;
 
+  /**
+   * Config options for navigation timing instrumentation. Disabled by default;
+   * set `enabled: true` to turn it on.
+   *
+   * Emits a `browser.navigation_timing` log record per page load, carrying the
+   * PerformanceNavigationTiming milestones.
+   */
```
wolfgangcodes, line 251:
```suggestion
   * Config options for navigation timing instrumentation. Disabled by default;
   * set `enabled: true` to turn it on.
   *
   * Emits a `browser.navigation_timing` log record per page load with the PerformanceNavigationTiming.
   */
```

MustafaHaddara (approved):

pkanal (approved):

PR #672, opened 2026-09-02
feat: add `shouldIgnoreError` option to global errors instrumentation

### LINKED ISSUE #666: Ability for the `GlobalErrorsInstrumentation` to skip errors.
in `@honeycombio/instrumentation-global-errors` applications have no way to skip errors they do not own, such as those thrown by browser extensions or third party scripts.

Could we have an option like `shouldIgnoreError` for the `GlobalErrorsInstrumentation` constructor function? The way I see it the function gets `error` object  as a argument. Applications can verify if this is a an error worth to be exported or not.

```TS
    shouldIgnoreError: (error) =>
      error.stack?.includes('chrome-extension://') ?? false,
  },
```

or make the `onError` property function accepts a `null` return that indicates to the instrumentation to skip this error.

The only way is to hook on the `onError` function, enable/disable the instrumentation based the error.

**Additional context**

tareqlol:
Applications have no way to skip errors they do not own, such as those thrown by browser extensions or third party scripts, out of their telemetry.

- Closes #666

Add a `shouldIgnoreError` option to the global error instrumentation that receives the error and the originating event so those errors can be filtered before a span is created.


tareqlol:
Hey team 👋🏼, 
Is there anything I can do to expedite this PR? Extension errors are creating too much noise on the honeycomb errors view page.

Thanks

wolfgangcodes:
As mention here: https://github.com/honeycombio/honeycomb-opentelemetry-web/issues/666#issuecomment-5647950255

We're not planning to expand the API at this time.

